PR #1, opened 2020-06-05
 openamp-xlnx: Parse non-dedicated IPIs for OpenAMP channels

bentheredonethat:
As OpenAMP uses channel information to configure IPC,
 add the follwing:

 - sample ipi's in input system DT
 - parsing for ipis in  openamp-xlnx assist
 - file generation in openamp-xlnx assist

### DIFF
--- a/assists/openamp-xlnx.py
+++ b/assists/openamp-xlnx.py
@@ -24,6 +24,52 @@ import importlib
 from lopper import Lopper
 from lopper import LopperFmt
 import lopper
+from lopper_tree import *
+
+openamp_file_name = "openamp_lopper_gen.h"
+
+# given interrupt list, write interrupt base addresses and adequate register width to header file
+# TODO append memory carveout-related info
+def generate_openamp_file(ipi_list):
+    f = open(openamp_file_name, "w")
+    # for each pair of ipi's present, write a master+remote ipi
+    f.write("#ifndef OPENAMP_LOPPER_INFO_H_\n")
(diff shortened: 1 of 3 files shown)


assists/openamp-xlnx.py
```diff
@@ -24,6 +24,52 @@
 from lopper import Lopper
 from lopper import LopperFmt
 import lopper
+from lopper_tree import *
+
+openamp_file_name = "openamp_lopper_gen.h"
```
zeddii, line 29: Ben, Do we think that the filename might vary ? In that case, this can be the default and we could check the options dictionary in the assist callback for a filename. That way it could passed via a lop file and changed if needed.

device-trees/system-device-tree.dts
```diff
@@ -54,6 +56,30 @@
 			};
 		};
 	};
+	ps_ipi_1 {
```
zeddii, line 59: I recall feedback that if there's a reg property in the node, the name should include the unit address, i.e. https://elinux.org/Device_Tree_Usage#Node_Names

If so, should we add some @<unit addresses> to these ?

bentheredonethat: added unit address to  name in later commit.

assists/openamp-xlnx.py
bentheredonethat: good point. added option to set file name via input r5 lop file in later commit

bentheredonethat:
Hi Bruce,

The latest commits should resolve your comments. 

Best Regards,
Ben

zeddii:
merged!

PR #3, opened 2020-06-10
openamp-xlnx: parse memory carveouts

bentheredonethat:
Output remoteproc and rpmsg compatible carveout information
into header for OpenAMP consumers

### DIFF
--- a/assists/openamp-xlnx.py
+++ b/assists/openamp-xlnx.py
@@ -25,11 +25,54 @@ from lopper import Lopper
 from lopper import LopperFmt
 import lopper
 from lopper_tree import *
+from re import *
 
+def write_one_carveout(f, prefix, addr_prop, range_prop):
+    f.write("#define ")
+    f.write(prefix+"ADDR\t"+addr_prop+"\n")
+    f.write("#define ")
+    f.write(prefix+"RANGE\t"+range_prop+"\n")
+
+def write_mem_carveouts(f, carveout_list, options):
+    symbol_name = "CHANNEL_0_MEM_"
+    current_channel_number = 0
+    channel_range = 0
+    current_channel_count = 0 # if == 4 then got complete channel range
+
+    for i in carveout_list:
+        if "channel" in i[0]:
+            # save channel number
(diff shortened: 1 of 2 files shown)


assists/openamp-xlnx.py
```diff
@@ -75,6 +122,19 @@ def parse_ipis_for_rpu(sdt, domain_node, rpu_cpu_node, options):
...
     return ipi_list
 
+def parse_memory_carevouts_for_rpu(sdt, domain_node, rpu_cpu_node, options):
+    try:
+        verbose = options['verbose']
+    except:
+        verbose = 0
+
+    carveout_list = []
+    for k,v in sdt.tree.__nodes__.items():
```
zeddii, line 132: In theory, we shouldn't be going directly as the __nodes__ member, since the implementation of it can change in the future (I'd have made it private if that was possible) ...

Have you tried this with the standard iterator ?

i.e.:

for node in sdt.tree:
   if channel in node.abs_path and "openamp,xlnx,mem-carveout" in node['compatible].value[0]:

... etc.

device-trees/system-device-tree.dts
```diff
@@ -81,6 +81,63 @@
 		reg = <0xFF370000 0x1000>;
 	};
 
+        channel0vdev0vring0: channel0vdev0vring0@3ed40000 {
+                no-map;
+		#address-cells = <2>;
+		#size-cells = <2>;
+                reg = <0x3ed40000 0x4000>;
```
zeddii, line 88: Is there a mix of tabs and spaces ? or is the alignment just displaying strangely ?

zeddii:
Sorry for the delay in looking through this, just the one comment on the iterator, and the possible tabs vs spaces.

device-trees/system-device-tree.dts
bentheredonethat: i apologize, it seems so. fixed in later commit.

assists/openamp-xlnx.py
bentheredonethat: ok this is now done as you suggested in later commit thank you

zeddii:
merged to master.

PR #4, opened 2020-06-18
assists: openamp-xlnx.py: output mem carveouts to r5 mem nodes

bentheredonethat:
As the openamp related memory carveouts as described in the input system DT and are output into the openamp specific header file, they should also be available for validation in output DTs. 

output the values to memory-region property in the corresponding R5 mem nodes to further this effort.


bentheredonethat:
> Did you want to go directly to .dtb, or through a .dts first ? If what comes out of lopper isn't valid for dtc, then that's a bug I need to fix.

Thanks for the clarification will update in next commit

zeddii:
merged

PR #5, opened 2020-06-23
assists: openamp-xlnx.py: add additional information for OpenAMP setup

bentheredonethat:
Add symbols using memory and interrupt specific info for the following
OpenAMP subsystems: VirtIO, RPMsg, Remoteproc.

Adding vector ID and bitmasks for versal IPIs, memory-related offsets and
trivial calculations and some hard-coded values corresponding to default
cases for resource table.

As the IRQ info is complex enough for 2 IPIs trim input systemDT to only have
needed default IPI info for now.

### DIFF
--- a/lops/lop-domain-r5-zynqmp.dts
+++ b/lops/lop-domain-r5-zynqmp.dts
@@ -1,0 +1,155 @@
+/*
+ * Copyright (c) 2019,2020 Xilinx Inc. All rights reserved.
+ *
+ * Author:
+ *       Bruce Ashfield <bruce.ashfield@xilinx.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/dts-v1/;
+
+/ {
+        compatible = "system-device-tree-v1,lop";
+        lops {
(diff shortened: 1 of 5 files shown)


bentheredonethat:
hi @zeddii  I will be adding at least 1 more commit for zynq as well so please do not merge yet

thank you!

lops/lop-domain-r5-zynqmp.dts
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright (c) 2019,2020 Xilinx Inc. All rights reserved.
+ *
+ * Author:
+ *       Bruce Ashfield <bruce.ashfield@xilinx.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
```
zeddii, line 8: There's a lot of extra stuff in this lops file, since it was driving a lot of testing early on. I'll merge this as-is, but then we can go through and drop all the extra bits in future commits.

zeddii:
I've merged 3 of the 4, with my change to address the 4th as part of the push.

PR #7, opened 2020-07-08
lopper.py: update -A option to check if files do not exist

bentheredonethat:
Raise FileNotFoundError if file is not found

This allows user to use lopper functionality within yocto recipe using only lopper and its static files as inputs - especially for openamp use cases

### DIFF
--- a/lopper.py
+++ b/lopper.py
@@ -595,9 +595,8 @@ class LopperSDT:
             print( "[DBG+]: assist_find: %s local search: %s" % (assist_name,local_load_paths) )
 
         try:
-            mod_file_abs = mod_file.resolve()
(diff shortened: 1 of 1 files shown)


zeddii:
I didn't merge this on purpose. It doesn't work with all versions of python, and I've already tested this locally and non existent files are flagged.

zeddii:
I should have added. If you send me your use case, I'll test against the versions of python I have available in the regression machines.

bentheredonethat:
the use case is building open-amp demos in yocto-land

e.g. using xilinx internal meta-openamp.

zeddii:
I've updated to python 3.6 locally and have tested this now. The change is now merged.

PR #8, opened 2020-07-16
Xilinx Baremetal driver assist files

kedareswararao:
Hi Bruce,

This patch series adds assists for xilinx baremetal use cases
Patch 
1/2: Add support for generating the baremetal driver config file.
2/2: Add support for returning valid driver list for provided system dt configuration.

Please review the changes and let me know your comments.

Regards,
Kedar.

### DIFF
--- a/assists/baremetalconfig_xlnx.py
+++ b/assists/baremetalconfig_xlnx.py
@@ -1,0 +1,304 @@
+#/*
+# * Copyright (c) 2020 Xilinx Inc. All rights reserved.
+# *
+# * Author:
+# *       Appana Durga Kedareswara rao <appana.durga.rao@xilinx.com>
+# *
+# * SPDX-License-Identifier: BSD-3-Clause
+# */
+import sys
+import types
+import os
+import getopt
+import re
+from pathlib import Path
+from pathlib import PurePath
+from io import StringIO
+import contextlib
+import importlib
+from lopper import Lopper
+from lopper import LopperFmt
+import lopper
+from lopper_tree import *
+from re import *
+import yaml
+
+
+def item_generator(json_input, lookup_key):
+    if isinstance(json_input, dict):
+        for k, v in json_input.items():
+            if k == lookup_key:
+                if isinstance(v, str):
+                    yield [v]
+                else:
+                    yield v
+            else:
+                for child_val in item_generator(v, lookup_key):
+                    yield child_val
+    elif isinstance(json_input, list):
+        for item in json_input:
+            for item_val in item_generator(item, lookup_key):
+                yield item_val
+
+# This API reads the schema and returns the compatible list
+def compat_list(schema):
+    if 'compatible' in schema['properties'].keys():
+        sch = schema['properties']['compatible']
+        compatible_list = []
+        for l in item_generator(sch, 'enum'):
+            compatible_list.extend(l)
+
+        for l in item_generator(sch, 'const'):
+            compatible_list.extend(l)
+
+        if 'contains' in sch.keys():
+            for l in item_generator(sch['contains'], 'enum'):
+                compatible_list.extend(l)
+
+            for l in item_generator(sch['contains'], 'const'):
+                compatible_list.extend(l)
+        compatible_list = list(set(compatible_list))
+        return compatible_list
+
+"""
+This API scans the device-tree node and returns the address
+and size of the reg property for the user provided index.
+
+Args:
+    node: LopperNode object
+    value: property value
+    idx: index
+"""
+def scan_reg_size(node, value, idx):
+    na = node.parent["#address-cells"].value[0]
+    ns = node.parent["#size-cells"].value[0]
+    cells = na + ns
+    reg = 0
+    size = 0
+    if cells > 2:
+        reg1 = value[cells * idx]
+        if reg1 != 0:
+            val = str(value[cells * idx + 1])
+            pad = 8 - len(val)
+            val = val.ljust(pad + len(val), '0')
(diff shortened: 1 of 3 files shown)


assists/baremetalconfig_xlnx.py
```diff
@@ -0,0 +1,292 @@
...
+    # Traverse the tree and find the nodes having status=ok property
+    for node in root_sub_nodes:
+        try:
+            status = node["status"].value
+            if "okay" in status:
+                node_list.append(node)
+        except:
+           pass
+
+    sdt.FDT = Lopper.dt_to_fdt(sdt.dtb, 'rb')
```
zeddii, line 181: You shouldn't need to do this. The sdt that is passed into the callback already has the dtb read, and assigned to sdt.FDT. That is the core system device tree.

assists/baremetalconfig_xlnx.py
```diff
@@ -0,0 +1,292 @@
...
+    driver_name = yamlfile.split('/')[-1].split('.')[0]
+    config_struct = str("X") + driver_name.capitalize() + str("_Config")
+    outfile = str("x") + driver_name + str("_g.c")
+
+    plat = DtbtoCStruct(outfile)
+    for index,node in enumerate(driver_nodes):
+        if index == 0:
+            plat.buf('#include "x%s.h"\n' % driver_name)
+            plat.buf('\n%s %s __attribute__ ((section (".drvcfg_sec"))) = {\n' % (config_struct, config_struct + str("Table[]")))
+        prop_dict = Lopper.node_properties_as_dict(sdt.FDT, node.abs_path)
```
zeddii, line 211: I had left the dictionary routines in the Lopper class as a transition. You can always have dictionary access to the properties using the tree nodes that are passed in. Fundamentally, we don't want fdt or other low level access in the assists, you can use the LopperTree methods exclusively.  If something is missing, it can be added.

assists/baremetalconfig_xlnx.py
```diff
@@ -0,0 +1,292 @@
...
+and size of the reg property for the user provided index.
+
+Args:
+    fdt: flattened device tree represention of the dts
+    node: dtb node
+    value: property value
+    idx: index
+"""
+def scan_reg_size(fdt, node, value, idx):
+    off = fdt.parent_offset(node)
```
zeddii, line 74: If you were to use the LopperTree nodes for this, finding the parent is simply: node.parent

assists/baremetalconfig_xlnx.py
```diff
@@ -0,0 +1,292 @@
...
+    na = Lopper.property_get(fdt, off, "#address-cells")
```
zeddii, line 75: And these properties would be: node["#address-cells"].value[0], etc.

zeddii:
Fundamentally, this looks good. I reviewed the one commit first, and then saw in the second one that you are aware of the LopperTree access methods. So I imagine that the generation commit is more of a direct port of how it used to work, and that's fine. So my comments there should only be seen as informational. 

We can worry about converting them to LopperTree format if the need comes, or if you have time, we can do some of it now.

This looks plumbed in properly, so if it is generating the code you need, then I don't have anything major to add.

If you do have a sample run, or test log, that would be useful, so I can run some simple integration tests to make sure I don't break it.

assists/baremetalconfig_xlnx.py
kedareswararao: I am getting sdt.FDT as null when my assist got called that's why i am trying to get the FDT again...

assists/baremetalconfig_xlnx.py
kedareswararao: Ok will update the code as you suggested and let you know if i found anything missing.

assists/baremetalconfig_xlnx.py
```diff
@@ -0,0 +1,292 @@
...
+            # interrupts: <supported range of interrupts>
+            if isinstance(prop, dict):
+               pad = list(prop.values())[0]
+               prop = list(prop.keys())[0]
+
+            if i == 0:
+                 plat.buf('\n\t{')
+
+            if prop == "reg":
+                val, size = scan_reg_size(sdt.FDT, sdt.FDT.path_offset(node.abs_path), prop_dict[prop], 0)
```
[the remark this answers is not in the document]
kedareswararao: Ok sure will update v2...

assists/baremetalconfig_xlnx.py
kedareswararao: ok sure will fix in v2...

assists/baremetalconfig_xlnx.py
kedareswararao: Sure will fix in next pull request

assists/baremetalconfig_xlnx.py
zeddii: Hmmm. That's almost impossible to happen, since things will immediately crash. Can you supply me a test command line ? That way I can debug it here ?

kedareswararao: Yes this line is not needed , i am not able to reproduce the issue that i have mentioned above, the issue that i observed may be due to a race in my script after cleanup things started working :) sorry for the noise.

kedareswararao:
Fixed review comments and force pushed changes in my tree, 
I am newbie to the pull request approach should i close this pull request and send one more pull request?

zeddii:
Either way works, adding/update commits or a new pull request. Since you have them on this existing one, I'll review them here and let you know if I have any questions.

PR #11, opened 2020-08-07
device-trees: remove amba_pl node as it is no longer present in BSP for testing

bentheredonethat:
amba_pl node interferes with OpenAMP IPC so remove this. also it is not used


zeddii:
merged.

PR #12, opened 2020-08-12
lops: lop-domain-r5-versal-rpmsg-kernelspace.dts: update for OpenAMP …

bentheredonethat:
…kernelspace case

Some nodes left in outputs interfere with memory mgmt for application. remove these so
there is no conflict


zeddii:
thanks ben. 

This is now merged.

PR #13, opened 2020-08-25
New Xilinx Baremetal Assists file and Bug fixes and Improvements 

kedareswararao:
1) Added assist file for generating baremetal board support package config
2) Added assist file for generating xparameters.h file
3) Renamed the linker assist file
4) Generate the required conf files for yocto tool

### DIFF
--- a/assists/baremetalconfig_xlnx.py
+++ b/assists/baremetalconfig_xlnx.py
@@ -23,6 +23,8 @@ from lopper_tree import *
 from re import *
 import yaml
 
+sys.path.append(os.path.dirname(__file__))
+from bmcmake_metadata_xlnx import *
 
 def item_generator(json_input, lookup_key):
     if isinstance(json_input, dict):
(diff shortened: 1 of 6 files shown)


assists/baremetalconfig_xlnx.py
```diff
@@ -203,24 +205,55 @@ def is_compat(node, compat_string_to_test):
         return xlnx_generate_bm_config
     return ""
 
+def get_stdin(sdt, chosen_node, node_list):
+    prop_dict = Lopper.node_properties_as_dict(sdt.FDT, chosen_node.abs_path)
```
zeddii, line 209: Is the chosen node not being properly pulled into the SDT already ? If not, that's an issue that I need to address.

We shouldn't need to go to the FDT at all to process parts of the tree.

zeddii:
Only one small comment from me. The assist cleanup and code generation look good.

If we can determine if there's a lopper bug based on my comment, or if we can change it to not read the FDT, I'll get the merge done immediately after that.

assists/baremetalconfig_xlnx.py
kedareswararao: I have fixed it and force pushed the change. Please take a look into it.

kedareswararao:
> Only one small comment from me. The assist cleanup and code generation look good.
> 
> If we can determine if there's a lopper bug based on my comment, or if we can change it to not read the FDT, I'll get the merge done immediately after that.

Thanks for the review @zeddii , I have fixed the bug and force pushed the change please review when you have some time and let me know your review comments 

Regards,
Kedar.

zeddii:
Looks good to me. This is now merged.

PR #14, opened 2020-09-03
Xilinx Baremetal Assists Feature Enhancments

kedareswararao:
Hi @zeddii ,

This patch series does the below
1) Added support for handling clocks and ranges property as per Baremetal requirements 
2) Updated the memory handling based on the CPU cluster address-map property
3) Added support for generating CCI define
4) Added support for memory tests app configuration.

Please review the changes when you have some time.

Regards,
Kedar.


zeddii:
merged!

PR #20, opened 2021-01-12
xlnx: add updated CDO for Versal Soc (1 board only)

bentheredonethat:
this patch does the following:
- re-enable cdo assist (presently xilinx versal support only)
- add lop file that:
  1. adds relevant domains
  2. for each domain, there are xilinx platform mgmt related nodes.
    a. associate these with the aforementioned domains
    b. for each node (device), ensure there is phandle for this node
       so it can be related to later
  3. in each domain, link each device to its specific xilinx platform mgmt
     flags
- retrofit cdo assist as follows:
  1. add lookup tables to hold xilinx-specific info
  2. parse domains, devices and the flags relating the devices to their domains
     and reflect this in output CDO

### DIFF
--- a/assists/cdo.py
+++ b/assists/cdo.py
@@ -26,7 +26,7 @@ import lopper
 from lopper_tree import *
 
 sys.path.append(os.path.dirname(__file__))
-import cdotypes
+from xlnx_versal_power import *
 
 def props():
     return ["id", "file_ext"]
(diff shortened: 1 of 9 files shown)


zeddii:
FYI: I haven't forgotten about this.  I've just been working through some other issues and will get to this shortly and give some review comments.

zeddii:
merged.  As planned, we can refine these further, but there's no risk in merging these initial versions.

PR #22, opened 2021-02-03
Update xilpm requirements

bentheredonethat:
remove old lop files and make more data driven approach from information propagated from lopper-input SDT


bentheredonethat:
i can clean up commit messages for this pull request

bentheredonethat:
updated commit messages and order of commits

bentheredonethat:
@zeddii  can you review this when able?

zeddii:
I have already scanned it once, and it looks ok.  

Lopper is going through an abstraction from libfdt and that has exposed some issues, and I've been dealing with that fallout.

Is there a full cut and paste line that you can provide for a sample run of lopper with these updated assists ? That way I can test against the new WIP code before merging.

zeddii:
There are still issues lurking due to the scope of the fdt abstraction changes, but I was able to run this and get the same output. Sanity is passing as well, so I've pushed this along with the fdt changes.

PR #27, opened 2021-02-26
assists: baremetalconfig_xlnx: Fix ranges and reg property greater th…

kedareswararao:
…an 32-bit address and size value handling

This commit fixes the race conditions in the greater than 32-bit value
handling for ranges and reg property.

Signed-off-by: Appana Durga Kedareswara rao <appana.durga.rao@xilinx.com>


zeddii:
merged

PR #28, opened 2021-03-10
Add support for Baremetal peripheral test app generation

kedareswararao:
1) Adds assist file for generating the baremetal peripheral test app
2) Add padding support for ranges property...


zeddii:
merged

PR #29, opened 2021-03-10
Add support for board topology cdo generation

nagasureshkumar:
This patch adds board topology cdo generation support to configure the
regulator rails.
We created a new lop file lop-versal-regulator-cdo.dts which contains
options argument as "regulator topology_board.cdo".
Here the options[0] which is regulator is to differentiate which cdo
the user is asking about and
the options[1] which is topology_board.cdo is the cdo file name.

The input to the lopper is system device tree which contains i2c
regulators information which will be parsed by cdo_topology.py.

The topology_headr.py contains all PM command and other macros required
to generated CDO data.

Signed-off-by: Naga Sureshkumar Relli <naga.sureshkumar.relli@xilinx.com>

### DIFF
--- a/assists/cdo.py
+++ b/assists/cdo.py
@@ -27,6 +27,7 @@ from lopper_tree import *
 
 sys.path.append(os.path.dirname(__file__))
 from xlnx_versal_power import *
+from cdo_topology import *
 
 def props():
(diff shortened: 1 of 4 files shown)


zeddii:
merged

PR #41, opened 2021-05-26
Cdo yaml support

bentheredonethat:
opening pull request now. the CDO plugin will be changed and cleaned up in the next week or two after some review of the design

### DIFF
--- a/assists/subsystem.py
+++ b/assists/subsystem.py
@@ -176,10 +176,7 @@ def access_expand( tree, subnode, verbose = 0 ):
     if not access_props:
         return
 
-    # the loop below used to use json.loads, which always returns a list,
-    # so we drop this in a list to avoid needing to change for the property
-    # direct access case.
-    access_chunks = [access_props[0][0]]
+    access_chunks = json.loads(access_props[0].value)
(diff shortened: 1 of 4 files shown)


assists/subsystem.py
```diff
@@ -216,60 +216,12 @@ def set_cdo_flags(flags, dev_node, dev_handle):
 
     return access_list
 
-def cdo_requirement_access_expand( tree, subnode, verbose , access_chunks ):
```
zeddii: So what's the overall change to access_expand here ? The standard access_expand is called from some simple domain processing, since that is part of the generic spec.

We should document the high level changes in the commit message itself, and indicate if the result is different for existing callers.

bentheredonethat: hi bruce,

im changing it right so there will be no more xilinx specific handling in the subsystem assist :)

ill update the pull request today

bentheredonethat:
@zeddii  the changes to subsystem expansion are less invasive on this branch now

zeddii:
Thanks for the updates. Reviewing now.

zeddii:
merged

PR #42, opened 2021-06-01
Add support for subsystem configuration for baremetal use cases

kedareswararao:
Add support for subsystem configuration for baremetal use cases...
@zeddii , When you have sometime please review the pull request and let me know the review comments...

Regards,
Kedar.

### DIFF
--- a/assists/baremetalconfig_xlnx.py
+++ b/assists/baremetalconfig_xlnx.py
@@ -417,6 +417,8 @@ def xlnx_generate_bm_config(tgt_node, sdt, options):
                if compat in compa:
                    driver_nodes.append(node)
 
+    # Remove duplicate nodes
+    driver_nodes = list(dict.fromkeys(driver_nodes))
     driver_nodes = get_mapped_nodes(sdt, driver_nodes, options)
     # config file name: x<driver_name>_g.c 
     driver_name = yamlfile.split('/')[-1].split('.')[0]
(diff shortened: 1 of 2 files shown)


zeddii:
I'll have a look shortly. I'm just finishing up some other urgent fixes and then will review.

assists/baremetalconfig_xlnx.py
```diff
@@ -309,7 +309,48 @@ def get_mapped_nodes(sdt, node_list, options):
...
+        try:
+            is_subsystem = node["cpus"].value
+            if is_subsystem:
+                domain_nodes.append(node)
+        except:
+            pass
+
+    if domain_nodes:
+        # Get Matched Domains for a given cpu cluster
+        cpu_phandle = Lopper.node_getphandle(sdt.FDT, match_cpunodes[0].parent.number)
```
zeddii, line 327: While this is ok, there are some incoming features that further split the lopper parser and backend from libfdt, including the introducing of a dtc-free run.  The Lopper class will contain the 'always available' methods, while front/backend specific ones will move to subclasses (and in this example, one for the FDT). 

If an assist is to work with all font/backends, it shouldn't directly reference the FDT (but instead, use the LopperTree routines).

In this case, you could simply access the parent.phandle, and if it is non-zero, you know the node has a valid phandle.

zeddii:
One comment on the use of the FDT, but otherwise, this looks fine.

assists/baremetalconfig_xlnx.py
kedareswararao: Thanks for the review... 
will fix and send the updated pull request...

kedareswararao:
@zeddii , Updated assist as per your review comments and force pushed the change, Please let me know if I need to send new pull request will close this and will send new one...

zeddii:
This same request is fine. I'll have a quick look today, and will get the merge done.

zeddii:
merged.

PR #43, opened 2021-06-11
Revert "lops: lop-domain-a72: in code block only remove bram. ipis ma…

nagasureshkumar:
…y be used"

This reverts commit 0e7d43a3ba9dc4ba3377c6434c629b1e909ac123.
Reverting this as Linux boot is hanging due to IPI nodes.
This patch is actually removes the IPI nodes that are not used by
Linux.

Signed-off-by: Naga Sureshkumar Relli <naga.sureshkumar.relli@xilinx.com>

### DIFF
--- a/lops/lop-domain-a72.dts
+++ b/lops/lop-domain-a72.dts
@@ -59,15 +59,35 @@
                 lop_8 {
                       compatible = "system-device-tree-v1,lop,code-v1";
                       code = "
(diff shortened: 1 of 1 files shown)


zeddii:
merged.

Also: note that there are significant refactoring changes on master now, so report any errors/issues and we'll get them fixed.

PR #49, opened 2021-06-26
RFC: convert lopper to python package

### LINKED ISSUE #30: setup lopper as a pypi package
Add whatever needs to be do such that lopper is available via pypi

mbolivar-nordic:
This relates to #30.

A prerequisite to making a lopper pypi package is having a lopper python package, in the sense of a lopper directory with an `__init__.py` etc.

This PR converts the lopper.py script and the various lopper_foo.py files into a lopper python package. Each lopper_foo.py is now lopper/foo.py. (This excepts lopper_sanity.py, which remains at top level.)

I'm posting this in incomplete form (e.g. I haven't tested, but I'm pretty sure I broke the vendored libfdt .so's) in hopes of getting feedback on whether you'd be open to the refactor.

If so, I can extend this to add pypi packaging files so people will be able to:

1. `pip3 install lopper`
2. run a `lopper` executable that gets inserted into PATH by `pip3` in step 1
3. `import lopper` and `import lopper.foo` from any Python file in the system

If not, I think we should probably just close #30, but I hope the benefits of having a pypi package outweigh the drawbacks of a painful reorganization.

Let me know; thanks!


mbolivar-nordic:
@zeddii I've updated this with most of the packaging boilerplate. I say "most" because I haven't found a good way to declare optional dependencies in setuptools. Would you be averse to just pulling in the kitchen sink (humanfriendly, flask, etc. etc.)?

Regarding backwards compatibility with running lopper.py as a script, I was able to do this:

```
$ cat bin/lopper.py
import sys
import subprocess

subprocess.run([sys.executable, '-m', 'lopper'] + sys.argv[1:])
$ pip install -e src/lopper   # this repository as of this pull request
$ python bin/lopper.py --help
... works as expected, prints lopper usage ..
```

Will that work?

zeddii:
> @zeddii I've updated this with most of the packaging boilerplate. I say "most" because I haven't found a good way to declare optional dependencies in setuptools. Would you be averse to just pulling in the kitchen sink (humanfriendly, flask, etc. etc.)?
> 
> Regarding backwards compatibility with running lopper.py as a script, I was able to do this:
> 
> ```
> $ cat bin/lopper.py
> import sys
> import subprocess
> 
> subprocess.run([sys.executable, '-m', 'lopper'] + sys.argv[1:])
> $ pip install -e src/lopper   # this repository as of this pull request
> $ python bin/lopper.py --help
> ... works as expected, prints lopper usage ..
> ```
> 
> Will that work?

Sorry I didn't get much of a reply yesterday, I'm still hunting a phandle resolution bug and didn't quite get to this.

But yes, that does look it should work. I can do some test runs to see how the existing/traditional users workflow will work and if they'll be supported.

It is funny that you ask about the optional dependencies, since I was just asked yesterday in another forum about making flask optional the same way that I've made other python dependencies optional since when integrating lopper into some build environments (i.e. OE) each dependency spawns a whole set of appended recipes and requires tweaks to work in a native/nativesdk (host/build server) side configuration.

So if there's not a good way to do that with this transformation, I'll ponder what that means for those types of installs.

zeddii:
I had to merge a few changes to fix some critical packaging and FDT writing issues. I realize that is a pain with this outstanding, since they will always conflict.

I'm now back to looking at this, and will take care of making sure it is rebased and valid for the tip of the tree.

I can sync up and help look into the optional dependencies now that the other issues have resolved (assuming they stay resolved).

mbolivar-nordic:
> I'm now back to looking at this, and will take care of making sure it is rebased and valid for the tip of the tree.

Sounds good; feel free to ping if there's anything else I can do of course.

PR #53, opened 2021-07-14
Update cdo parsing

bentheredonethat:
- yaml -> dts now matches spec
- update cdo plugin to use latest generated SDT per YAML expansion

### DIFF
--- a/assists/cdo.py
+++ b/assists/cdo.py
@@ -29,7 +29,6 @@ sys.path.append(os.path.dirname(__file__))
 from xlnx_versal_power import *
 from cdo_topology import *
 from collections import OrderedDict
-from subsystem import *
 import json
 import humanfriendly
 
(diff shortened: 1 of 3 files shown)


zeddii:
I've been having a look at this, and want to consider a few more things. 

I wanted to comment, so you'd know that I didn't miss this!

bentheredonethat:
@zeddii  ping for update

zeddii:
Still the same as 2 days ago. reviewing and testing. There are interactions with other pending changes.

zeddii:
I'm finishing up a generalized description language that would simplify some of the assists by allowing us to describe the yaml flags -> DTS flags, without needing to write as much python code.

But that is taking a bit longer than I'd like, so I've gone ahead and merged this.

We can adjust as necessary once I get that code completed.

PR #59, opened 2021-07-29
Openamp parsing

bentheredonethat:
- work with latest SDT
- scale memory expansion in YAML to work with lists (used in openamp parsing later)
- add flag to denote  kernel/userspace flows
- add flag to denote  host/remote in an OpenAMP channel
- in userspace flow, have host IPI 'interrupts' set dynamically from referenced host IPI
- in kernelspace flow, generate mailbox controller based on included IPIs from host and remote
- simplify memory carveout construction


assists/subsystem.py
```diff
@@ -264,7 +264,7 @@ def memory_expand( tree, subnode, memory_start = 0xbeef, prop_name = 'memory', v
     # * memory = <address size address size ...>
     # */
     try:
-        mem = [subnode.props( prop_name )[0][0]]
+        mem = json.loads( subnode.props ( prop_name)[0].value )
```
zeddii, line 267: We don't want explicit json loads called within this assist. It needs to be at a more generic level.

This is undoing an explicit change that removed all the json.loads() calls, so we can't put them back in.

The LopperProperty can already automatically do the json loads if the pclass is correct, and it can return the len of the json property.

so we should be doing something like:

mem = []
prop = subnode[prop_name]
for i in range(0,len(prop)):
    mem.append( prop[i] )

... and then iterate on m, which are the json loaded chunks.

zeddii:
Everything looks fine, except for the subsystem.py change.

If the LopperProperty json routines are failing, raise an issue here, and I'll fix those, but we don't want explicit json.loads() calls.

bentheredonethat:
Hi @zeddii,

I'll check on whether its LopperProperty json routines or something on my end.

Thanks!

bentheredonethat:
@zeddii  updated PR per review

zeddii:
This looks fine to me. I've queued it.

There are some yaml expansion issues I'm seeing, but if I can't get them sorted within then next day or so, I'll rebase and merge this.

PR #60, opened 2021-08-09
Support for domain/device level protections (xlnx)

its-izhar:
- Support for domain level protections (firewallconf/firewallconf-default)
 - Support for device level protections through firewall table (prot.config/prot.config.in)
 - Marker support for subsystems/protection config in output CDO
 - Cleanup top level CDO plugin and move xilinx platform-management specific code to separate area in assists (assists/xlnx_pm/*)
 - Formatting changes and cleanups


zeddii:
I'm now back from travelling (but not fully back yet), and will do a more detailed review shortly (Friday or Monday) as I update my other pending changes (firewallconf, etc).

zeddii (commented):
Was the formatting / cleanup done via a linter/pep8/other utility ?

If so, we need to document that in the commit header, so we can ensure consistency in the future.

zeddii (commented):
Ignore this comment, it was dissociated with the commit in question.

zeddii (commented):
While there's nothing structurally wrong with this .. subsystems are not only for xilinx_pm (or cdo)

If we are doing a cleanup, we should put these in a xlnx directory, but not specifically xlnx_pm

assists/cdo.py
```diff
@@ -1,4 +1,4 @@
-#/*
+# /*
```
zeddii, line 1: Was the formatting / cleanup done via a linter/pep8/other utility ?

If so, we need to document that in the commit header, so we can ensure consistency in the future.

assists/cdo.py
```diff
@@ -51,144 +51,6 @@ def is_compat(node, compat_id):
     return ""
 
 
-def do_expand_cdo_flags(tgt_node):
```
zeddii, line 51: Can we tweak the commit message to indicate why this is being done ? Is it in preparation for future changes ? cleanup ? import simplification ? something else ?

assists/cdo.py
```diff
@@ -29,6 +29,7 @@
 from lopper import Lopper
 import lopper
 from lopper_tree import *
+from xlnx_pm import subsystem
```
zeddii, line 32: While there's nothing structurally wrong with this .. subsystems are not only for xilinx_pm (or cdo)

If we are doing a cleanup, we should put these in a xlnx directory, but not specifically xlnx_pm

zeddii:
The series looks good. My only comments are on the first few commits, and the only significant one being a tweak to the restructuring.

But I like the result of this, much cleaner and easier to read.

I haven't reviewed the cdo/pm commits in detail, except to look and ensure that no obsolete or depreciated device tree / Lopper access routines were being used.

assists/cdo.py
its-izhar: okay, sure will document the utility (or format) in the commit header.

I used "yapf" (https://github.com/google/yapf) - by default it does pep8 formatting, which is what I used. It supports other formats however.

assists/cdo.py
its-izhar: The reasoning for this is mainly code cleanup and better code organization.

do_expand_cdo_flags() and other helpers are  mainly doing the YAML to DTS flags expansion for domains/subsystems. This code correctly belongs in assists/subsystem.py (xlate-yaml) plugin given its role in the subsystem expansion.
That's why this change.

I will update the commit message to add these details.

assists/cdo.py
its-izhar: Initially, I thought of going with "xlnx" instead of "xlnx_pm".
But I am thinking if we do "xlnx", people may start adding other non-cdo generation related xilinx plugins in that directory. I intended this folder to only be limited to assists for cdo generation with subsystems and protections.

Let me know what do you think, I am fine either way.

its-izhar:
> 
> 
> The series looks good. My only comments are on the first few commits, and the only significant one being a tweak to the restructuring.
> 
> But I like the result of this, much cleaner and easier to read.
> 
> I haven't reviewed the cdo/pm commits in detail, except to look and ensure that no obsolete or depreciated device tree / Lopper access routines were being used.

Thanks for the review. One of the goals I was trying to achieve with this was to improve readability and code organization - happy that it is meeting your expectations.

I will incorporate the feedback - replied to each comment with explanation and my thoughts. Let me know your thoughts about the xlnx vs xlnx_pm dir names.

assists/cdo.py
zeddii: Sounds good. The plan was to do pep8, so that is the preference (for now).

assists/cdo.py
zeddii: perfect. Thanks!

assists/cdo.py
zeddii: We can always have both. xlnx and xlnx_pm, even if there is very little in xlnx initially.  subsystem in particular has the potential to be used for more than just plm/cdo and even non xilinx applications, so having it in xlnx, prepares it for extension / migration in the future.

its-izhar: Okay, sounds good. I will use xlnx for now.

its-izhar:
Closing this pull request as a new one is opened with comments resolved and a couple more features added.

Thanks!

PR #61, opened 2021-08-10
cdo: update parsing to include QOS param (param 4 starting from index 0)

bentheredonethat:


its-izhar:
@bentheredonethat Can you send me this patch, I will include it as a part of my open pull-request (#60)

This will avoid unnecessary merge conflicts. I have a lot of re-factoring in there so I can add this easily. Let me know.

PR #62, opened 2021-08-12
assist: Update the zuplus_config with latest system device tree spec

ganna09:
There are changes in the latest system device tree subsystem specification
like access and memory properties.
Ex: previously the access = <&can_1 0x1>;
now access = <&can_1>; and flags are given under a separate property
named flags. Same applies to memory property.

Signed-off-by: srilaxmi <srilaxmi@xilinx.com>


zeddii:
Although the FDT (libfdt) calls were in the previous versions of the assist, we should take the opportunity to remove them, and support the non-libfdt modes (which are required for on-target and python only use cases). I'll close this, and wait for an updated pull request.

PR #63, opened 2021-08-20
assists: baremetal*: Remove references to sdt.FDT handling

kedareswararao:
Inorder to support PDT (parsed device-tree) SDT.FDT calls
shouldn't be used in the assist files.

This commit updates the assist for the same.

@zeddii : After this commit still we are using sdt.FDT calls at two places in baremetal assists,
1) sdt.FDT.get_alias (Couldn't able to find equivalent API in lopper tree)
2) Lopper.node_properties_as_dict(sdt.FDT) (This can be resolved need to look into lopper front end..)

### DIFF
--- a/assists/baremetal_xparameters_xlnx.py
+++ b/assists/baremetal_xparameters_xlnx.py
@@ -144,7 +144,7 @@ def xlnx_generate_xparams(tgt_node, sdt, options):
                             canondef_dict.update({prop:node[prop].value[0]})
                         elif prop == "interrupts":
                             try:
-                                intr = get_interrupt_prop(sdt.FDT, node, node[prop].value)
+                                intr = get_interrupt_prop(sdt, node, node[prop].value)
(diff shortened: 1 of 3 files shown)


zeddii:
This is a good first step!

You are correct, there isn't currently an abstracted alias interface, but I can add one. Do you have a sample input device tree, and baremetal assist command line that I can use for testing ? That way I can be sure the interface does what you need.

As for the second one. Pretty much every LopperNode is already a dictionary. You can access any parameter via Node["property_name"], which is your equivalent dictionary. You can also iterate the properties using a standard iteration loop.

So rather than creating a new interface for dictionary export, that would the preferred access mechanism to use.

But that switch won't hold up this pull request, I'll get it queued and we can do the alias/dictionary changes in the future.

zeddii:
I went ahead and created a mapping for node aliases in LopperTree. See the newly added "alias_node" method. The details are in the commit message at the head of master.

This is passing sanity tests, but we can tweak it more if it doesn't do exactly what you need.

PR #64, opened 2021-08-20
 Support for domain/device level protections (xlnx) [v2]

its-izhar:
- Support for domain level protections (firewallconf/firewallconf-default)
- Support for device level protections through firewall table (prot.config/prot.config.in)
- Marker support for subsystems/protection config in output CDO
- Cleanup top level CDO plugin and move xilinx platform-management specific code to separate area in assists (assists/xlnx_pm/*)
- Formatting changes and cleanups
- Added support for memory protections

Resolved following in the revew comments from [earlier pull request:](https://github.com/devicetree-org/lopper/pull/60)
- Use 'xlnx' instead of 'xlnx_pm' for under assists/* for code organization
- Add formatting information (pep8) to commit message
- Enhance commit message for moving yaml -> dts flags expansion inside xlate-yaml


its-izhar:
Earlier pull request with comments (in case needed for reference): https://github.com/devicetree-org/lopper/pull/60

PR #69, opened 2021-09-08
assists: baremetal*: Bug fixes

kedareswararao:
1) Fix sdt.FDT checks in baremetal assist files 
2) Fix race condition in baremetal linker assist for memory test
3) Update A72 domain lops file as per latest system device-tree generated by the device-tree-generator tool 
4) Update A53 linux domain lops file as per latest system device-tree generated by the device-tree-generator tool

### DIFF
--- a/assists/baremetal_xparameters_xlnx.py
+++ b/assists/baremetal_xparameters_xlnx.py
@@ -35,7 +35,7 @@ def is_compat( node, compat_string_to_test ):
     return ""
 
 def get_label(sdt, symbol_node, node):
-    prop_dict = Lopper.node_properties_as_dict(sdt.FDT, symbol_node.abs_path, False)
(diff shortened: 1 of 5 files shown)


zeddii:
Just one comment / question on the label lookup from me, otherwise, this looks fine.

zeddii:
> Hi @zeddii ,
> Thanks, If you don't mine could you please apply the other patches in the pull request if you are ok with the changes.
> 
> Regards,
> Kedar.

I'll merge the pull request as-is, but I also just pushed a thin wrapper in LopperNode to abstract what you want. You can now just call items() on the node itself, and it returns the items() from the internal __props__ attribute.

As for the use case, the label of the node wouldn't return what you want, due to the "@" processing (as you provided).

PR #85, opened 2021-10-11
assists: baremetal_xparameters_xlnx: Generate canonical defines for generic nodes

kedareswararao:
This pull request updates the assist to generate canonical defines generic nodes,

@zeddii : Please apply this commit to only master branch...


assists/baremetal_xparameters_xlnx.py
```diff
@@ -232,14 +232,33 @@ def xlnx_generate_xparams(tgt_node, sdt, options):
                                     
     # Generate Defines for Generic Nodes
     node_list = get_mapped_nodes(sdt, node_list, options)
+    prev = ""
+    count = 0
     for node in node_list:
+        prop_dict = node.__props__
```
zeddii, line 238: prop_dict doesn't seem to be used.

I noticed it immediately, since as you know, we try to avoid going directly at this dictionary.

Is it used somewhere outside of the diff ? or am I missing something else ?

PR #87, opened 2021-10-18
domain support in barematal and linux domains

kedareswararao:
This pull request adds domain and resource handling support in baremetal and linux domains,
Existing logic handles domains based on the os,type property.

@zeddii : 
I have addressed most of the review comments and rebased the changes on top of master-next branch, 
Please review the changes when you have sometime.

Regards,
Kedar.

### DIFF
--- a/lopper/lops/lop-domain-a72.dts
+++ b/lopper/lops/lop-domain-a72.dts
@@ -42,11 +42,54 @@
                       compatible = "system-device-tree-v1,lop,select-v1";
                       // clear any old selections
                       select_1;
+                      select_2 = "/domains";
+                };
+                lop_5_0 {
+                      compatible = "system-device-tree-v1,lop,code-v1";
+                      code = "
+                          assist_dir = os.path.dirname(os.path.realpath(__file__)) + '/assists/'
+                          sys.path.append(assist_dir)
+                          from domain_access import update_mem_node
+                          from baremetalconfig_xlnx import scan_reg_size
+                          // Domain memory handling
+                          shared_mem = []
+                          memnode_list = tree.nodes('/memory@.*')
+                          for n in node.subnodes():
+                              if n.propval('memory') != [''] and n.propval('os,type') == ['']:
+                                  shared_mem.append(n)
+                          for n in node.subnodes():
+                              if n.propval('memory') != [''] and  n.propval('os,type') != ['']:
+                                  if re.search('linux', n.propval('os,type', list)[0]):
+                                      start_addr = n.propval('memory')[0]
+                                      linux_shared_mem = []
+                                      //Check for resource group
+                                      if n.propval('include'):
+                                          // Get shared memory nodes for Linux domain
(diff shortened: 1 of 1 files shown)


lopper/assists/domain_access.py
```diff
@@ -40,6 +40,65 @@ def check_bit_set(n, k):
...
+    return memory_nodes
+
+# node: is the domain node number
+# mem_val: Memory node address and size value to be updated
+# This api takes the memory value(address and size) and creates
+# a new memory node value for memory node reg property based on the
+# address-cells and size-cells property.
+def update_mem_node(node, mem_val):
+    ac = node.parent['#address-cells'].value[0]
+    sc = node.parent['#size-cells'].value[0]
```
zeddii, line 85: If you want to protect yourself from inputs that come via yaml/other formats (not that it is common), and also abstract the property a bit, these calls can also just be:

ac = node.parent['#address-cells'][0]

And let the smart indexing get you the first value. (since you aren't looking to have value as a list in your use case anyway).

lopper/assists/baremetallinker_xlnx.py
```diff
@@ -30,19 +30,72 @@
...
+                    bm_domain.append(domain_node)
+        elif domain_node.propval('cpus') != ['']:
+                print('ERROR: os,type property is missing in the domain', domain_node.name)
+
+        match_cpunodes = get_cpu_node(sdt, options)
+
+        # Check whether domain is mapped for current processor or not
+        if bm_domain:
+            shared_mem = []
+            if not domain_node['cpus'].value[0] == match_cpunodes[0].parent.phandle:
```
zeddii, line 64: Minor comment: Similar to my other comment, this could just be domain_node['cpus'][0]

lopper/assists/baremetallinker_xlnx.py
```diff
@@ -30,19 +30,72 @@
...
+                continue
+            else:
+               if domain_node.propval('include') != ['']:
+                   rsc_domain = domain_node['include'].value
```
zeddii, line 68: There's already a lopper_lib routine called 'includes' that processes an include property and returns all of the target nodes.

We should use that here (and fix / tweak the library option if it doesn't work).

i.e. lopper_lib.includes( tree,  domain_node['include'] )

Should chase down the target of the include.

lopper/lops/lop-domain-linux-a53.dts
```diff
@@ -40,6 +40,7 @@
                       select_2 = "/cpus/.*:compatible:.*arm,cortex-a53.*";
                       select_3 = "/.*:status:.*okay.*";
                       select_4 = "/.*:device_type:.*memory.*";
+                      select_5 = "/domains";
```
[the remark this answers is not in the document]
kedareswararao: @zeddii , Thanks for the review the problem here is we need both domains related data and status ok nodes for processing, that's why implemented like this other wise there will be duplicate code if i split the code as you suggested... 

If duplicate code is ok please let me know will update the code as you suggested

PR #102, opened 2022-03-30
Openamp process tree

bentheredonethat:


bentheredonethat:
@zeddii  this shows Versal and ZU+ OpenAMP Xilinx domain processing

zeddii:
I'm just finishing up some other changes/testing, I'll have a look at this shortly.

lopper/lops/lop-domain-a72-prune.dts
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright (c) 2020 Xilinx Inc. All rights reserved.
```
zeddii, line 2: Minor 'nit and no need to change this for this pull request, but this should be updated 2022 for new files, even if we are just moving things around.

lopper/assists/openamp_xlnx.py
```diff
@@ -743,6 +743,9 @@ def xlnx_remoteproc_construct_cluster(tree, cpu_config, elfload_nodes, new_ddr_n
         for carveout in elfload_nodes:
             if carveout.props("status") != []:
                 srams.append(carveout.phandle)
+                # FIXME for each sram, add 'power-domain' prop for kernel driver
+                carveout + LopperProp(name="power-domain",
+                                      value=copy.deepcopy( carveout.props("power-domains")[0].value ))
```
zeddii, line 760: Note: if you want the entire property to be deep copied for a new one,  LopperProp (and LopperNode) implement the __call__ magic function (and do a deepcopy by default). So you can do new_prop = old_prop()

But this is ok, as is .. that is just for reference.

lopper/assists/openamp_xlnx.py
```diff
@@ -632,6 +632,14 @@ def xlnx_rpmsg_parse(tree, node, openamp_channel_info, verbose = 0 ):
...
 
 
+# tests for a bit that is set, going fro 31 -> 0 from MSB to LSB
+def check_bit_set(n, k):
+    if n & (1 << (k)):
+        return True
+
+    return False
+
+
```
zeddii, line 654: This is also available in lopper_lib, so you can import it from there. A potential cleanup idea for future changes.

zeddii:
only minor comments on this, it won't impact the merge. They are just for reference.

PR #103, opened 2022-04-08
Small bug fixes a53 linux

bentheredonethat:


zeddii (commented):
What happens if the node is present in the tree ? Will there be an issue ? We can make this more robust, if we need to conditionally remove the node.

bentheredonethat:
ok will send an update to make this more robust @zeddii

bentheredonethat:
@zeddii  updated to make this more robust. please review when able

PR #110, opened 2022-06-14
Revert "lopper/tree: improve phandle and node operations"

bentheredonethat:
This reverts commit 5ae66502ea98dfa7527d21b34796e07904f30a6a.

this commit causes breakages for openamp lopper runs

@zeddii

### DIFF
--- a/lopper/tree.py
+++ b/lopper/tree.py
@@ -77,8 +77,6 @@ class LopperProp():
         self.ptype = ""
         self.binary = False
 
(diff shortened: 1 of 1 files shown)


zeddii:
Sorry, no. We can't revert that commit. It is required functionality for many other features.

I'll have a look at the openamp issues ASAP, but cycles are limited for the next few days.

You really need to prepare something that is fully integrated with the lopper_sanity script, since that is run before anything is pushed.

PR #119, opened 2022-09-02
extract-xen: Handle xen,reg properly

orzelmichal:
Currently one of the major issue with extract-xen assist is that it
hardcodes 2nd/3rd groupings which means that xen,reg property is setup
incorrectly for nodes with reg property containing more than one memory
region. Add a proper handling to fix this issue.

Signed-off-by: Michal Orzel <michal.orzel@amd.com>


zeddii:
merged!

PR #120, opened 2022-10-03
fdt: Fix for the condition when an include file is defined inside a node

onkarharsh:
When an include file is defined inside a node as shown below, lopper is throwing an error.

&qspi {
#include "...dtsi"
};

Whenever a line starting with # is detected, subnode_at_depth is re-initialized to 0 and False, but the key that is being used to traverse through subnode_at_depth i.e. node_depth never gets decremented in above situation.

Signed-off-by: Onkar Harsh <onkar.harsh@xilinx.com>


onkarharsh:
Closing this PR due to the wrong signed off statement. Will open a new PR.

PR #121, opened 2022-10-03
fdt: Fix for the condition when an include file is defined inside a node

onkarharsh:
When an include file is defined inside a node as shown below, lopper is throwing an error : key (node_depth) out of bound of the dictionary (subnode_at_depth). 

&qspi {
#include "...dtsi"
};

Whenever a line starting with # is detected, subnode_at_depth is re-initialized to 0 and False, but the key that is being used to traverse through subnode_at_depth i.e. node_depth never gets decremented in above situation.

CC: @zeddii @kedareswararao


zeddii:
This passed sanity testing, and looks ok to me.

Thanks for the fix, this is now merged to master.

PR #122, opened 2022-10-10
lops/assists: Changing the bus name from amba to axi in all the lops/assists.

onkarharsh:
As per the upstreamed linux syntax, "amba" is not a valid bus name and the bus name "amba" has been replaced with "axi". Reference linux commits can be found below. This patch brings the same change inside lopper to keep the device trees in sync with the linux nomenclature.

Reference Linux commits:
https://github.com/torvalds/linux/commit/dfff9066e60ee8670424cfedf4269d3b51c5100f https://github.com/torvalds/linux/commit/74790cf981afd9806939427ad7701f3022ee5c06

CC: @zeddii @kedareswararao @sivadur


zeddii:
To make this complete, we really need to look for all of the amba references and change them as well.

I see some of the docs, specification and reference trees that also have amba references. We can ignore the lopper_sanity references, as they are just unit tests.

onkarharsh:
Thanks @zeddii. Made the respective changes in all the docs and specs.

zeddii:
This version looks good. I'm running some sanity tests on it, and will merge once they pass.

onkarharsh:
Hi @bentheredonethat / @zeddii ,
From regressions report with above change, I confirm that all the open-amp demos mentioned above work.

I guess, we can now proceed to merge the patch.

PR #123, opened 2022-11-09
lopper: assits: Improvements in baremetal assists

kedareswararao:
1) Add support for outdir option in assists 
2) Added new assist for getting the supported apps and libs list for supported os along with it's description 
3) Added new lops file getting cpu names for a given system device-tree
4) Updated the assists to remove hard coded cpu compatible to yocto machine dictionary and fixed assists to take cpu name as input instead of yocto machine name.


zeddii:
I haven't missed this, just haven't done a full review yet. Will start that shortly.

PR #126, opened 2022-11-23
specification: convert to Sphinx

mbolivar-nordic:
Convert the core system devicetree specification to the restructuredtext file format (.rst) and Sphinx documentation system.

This is the same documentation system used by both the (regular) devicetree specification and the Linux documentation.

Using this format over plain markdown has many advantages. A few are:

- auto-numbered sections
- HTML output
- PDF output (via LaTeX)
- the ability to easily reference other sections
- syntax highlight for DTS fragments

Include a README.md for how to build the source. To avoid bitrot and skew, delete the .md documents which were converted to the new format.

I took care to make the minimum number of formatting changes possible during this conversion, avoiding typo fixes, unnecessary formatting changes, etc. This is meant as a starting point for further refinements where each .rst page should be easy to compare to its starting point.

I left a few documents out of the conversion when it wasn't clear that they were part of the 'core' system DT spec. If that's wrong and they are meant to be part of the spec somehow, that's easy enough to fix as well.


mbolivar-nordic:
Here is a PDF:

[system-devicetree-specification.pdf](https://github.com/devicetree-org/lopper/files/10072602/system-devicetree-specification.pdf)

And here is a .zip of the HTML output (all in one page -- "one page per chapter" output is also available; see README.md):

[singlehtml.zip](https://github.com/devicetree-org/lopper/files/10072614/singlehtml.zip)

(GitHub doesn't allow .tgz, sorry.)

zeddii:
I can't easily tell if the content is exactly the same in the before an after, given that some files are moved around.

That being said, what is there is still readable when looking at the source, so I have no issues and I assume that everything is the same content wise.

Once I've run a few experiments on building the docs locally, I'll get this into the tree.

zeddii:
I was able to build the docs, pushed to master.

Issue #130, opened 2022-11-24
Purpose of secure-bus?


mbolivar-nordic:
When I was last working on system devicetree, we were discussing how to represent address maps for Arm v8-M CPUs with TrustZone support. In this case, the same IP blocks often effectively have different addresses depending on the security state of the CPU at the time of access.

It looks like secure-bus might be meant as a solution to that. Is that so? If so, I think it's not needed at least for Zephyr and it actually complicates things for us. I think having as many `address-map` properties as needed, with identical semantics (and therefore no differences in the places where the number of address and size cells are set; see https://github.com/devicetree-org/lopper/issues/129 for more on this) would be better.

In effect this is similar to the way the `pinctrl-<N>` properties work for pin control: any node with pin control support can have `pinctrl-0`, `pinctrl-1`, ... properties, as many as needed, and with identical semantics. Each property represents a different pin selection depending on the state of the IP block modeled by the node.

Similarly, as many address map properties as needed, each of which reflects some state of the cpus cluster, would seem to work well for us handling Arm v8-M in zephyr, at least for now. (This opens up the possibility of having an `address-map` or `address-map-0` property, then `address-map-1`, `address-map-2`, ... as well, but that is probably a separate issue.)

zeddii:
@sstabellini

sstabellini:
> When I was last working on system devicetree, we were discussing how to represent address maps for Arm v8-M CPUs with TrustZone support. In this case, the same IP blocks often effectively have different addresses depending on the security state of the CPU at the time of access.
> 
> It looks like secure-bus might be meant as a solution to that. Is that so? If so, I think it's not needed at least for Zephyr and it actually complicates things for us. I think having as many `address-map` properties as needed, with identical semantics (and therefore no differences in the places where the number of address and size cells are set; see #129 for more on this) would be better.
> 
> In effect this is similar to the way the `pinctrl-<N>` properties work for pin control: any node with pin control support can have `pinctrl-0`, `pinctrl-1`, ... properties, as many as needed, and with identical semantics. Each property represents a different pin selection depending on the state of the IP block modeled by the node.
> 
> Similarly, as many address map properties as needed, each of which reflects some state of the cpus cluster, would seem to work well for us handling Arm v8-M in zephyr, at least for now. (This opens up the possibility of having an `address-map` or `address-map-0` property, then `address-map-1`, `address-map-2`, ... as well, but that is probably a separate issue.)

[...]

Issue #128, opened 2022-11-24
Incorrect usage of unit addresses in various nodes


mbolivar-nordic:
From the devicetree specification, section 2.2.2 Node Names:

> The
> unit-address must match the first
> address specified in the *reg* property of the node. If the node has no
> *reg* property, the *@unit-address* must be omitted and the
> *node-name* alone differentiates the node from other nodes at the same
> level in the tree.

From this we can also infer a derived requirement that any node with a unit address must have a parent node with `#address-cells` set, and the `reg` property must be consistent with it.

The system devicetree specification and associated examples violates this requirement in several places:

1. system-device-tree.dts

 https://github.com/devicetree-org/lopper/blob/da29be7393b1b3237c3a566b42a5dd5f81cf391c/specification/system-device-tree.dts 

This has a root node with `#address-cells = <2>`, so the `indirect-bus` and `cpus,cluster` child nodes should either 1. not have unit addresses or 2. have `reg` properties with 2 cells for the address. This is also true of multiple `simple-bus` nodes which are children of the root node.

2. specification.md 

https://github.com/devicetree-org/lopper/blob/da29be7393b1b3237c3a566b42a5dd5f81cf391c/specification/specification.md

Same issue: various nodes have unit addresses but no `reg` property. Additionally, the context makes it unclear whether the parent node has `#address-cells` set.

zeddii:
The captured reference tree needs some updating, it has been on the TODO list for a while. There are some other sample device trees floating around, but they are vendor specific, so I haven't wanted to directly use them.

We need to get @sstabellini to comment on this (hopefully tagging him here works)

sstabellini:
> From the devicetree specification, section 2.2.2 Node Names:
> 
> > The
> > unit-address must match the first
> > address specified in the _reg_ property of the node. If the node has no
> > _reg_ property, the _@unit-address_ must be omitted and the
> > _node-name_ alone differentiates the node from other nodes at the same
> > level in the tree.
> 
> From this we can also infer a derived requirement that any node with a unit address must have a parent node with `#address-cells` set, and the `reg` property must be consistent with it.
> 
> The system devicetree specification and associated examples violates this requirement in several places:
> 
>     1. system-device-tree.dts
> 
> 
> https://github.com/devicetree-org/lopper/blob/da29be7393b1b3237c3a566b42a5dd5f81cf391c/specification/system-device-tree.dts
> 
[...]

Issue #129, opened 2022-11-24
#ranges-address-cells and #ranges-size-cells are undefined


mbolivar-nordic:
These properties are not defined in specification.md:

https://github.com/devicetree-org/lopper/blob/da29be7393b1b3237c3a566b42a5dd5f81cf391c/specification/specification.md

They appear to be related to the address-map property, but they are not defined in the specification and their purpose is unclear. They only appear in the system-device-tree.dts document, whose relation to the specification is not clear, because specification.md never refers to system-device-tree.dts.

It is also unclear why the secure-address-map property should have a different location where the number of address and size cells is set than the regular address-map property, which uses #address-cells and #size-cells as far as I can tell from examples. But again, this is not defined in the specification.

zeddii:
@sstabellini this is for you (but i can't assign it, as you aren't in the project yet).

sstabellini:
Let's put aside secure-address-map given that it is about to go away as per issue #130.

As you can see from this patch:
https://lists.openampproject.org/archives/list/system-dt@lists.openampproject.org/thread/QEQAZUXYKT7P37RBSD4LBR5KRNWRUCVF/

#ranges-address-cells represents the number of cells to use for the first address in address-map.
#ranges-size-cells represents the number of cells to use for the size (last parameter) in address-map.

I realize that the detailed description of the binding as per the patch above didn't make it into specification.md. If you are happy to add that to the pull request, we'll merge it.

mbolivar-nordic:
Thanks for the reference!

> As you can see from this patch: https://lists.openampproject.org/archives/list/system-dt@lists.openampproject.org/thread/QEQAZUXYKT7P37RBSD4LBR5KRNWRUCVF/

I am unaware of the repository this patch applies to, though -- could you please explain what this is? My concern here is that there seem to be two places where system devicetree related content is being defined. IMO it would be preferable for core concepts like address map to be defined in the specification, and I'm worried about bitrot and content skew if there are multiple places where definitions are going.

sstabellini:
Sorry for the confusion: that patch was never applied anywhere and predates the System Device Tree specification in the Lopper repository. Today there is only one place for the System Device Tree specification and it is under the Lopper repository.

However I linked the patch because it has a good write-up to define address-map together with #ranges-address-cells and #ranges-size-cells which is missing in specification.md. We need to add that content to specification.md in the Lopper repository.

mbolivar-nordic:
[...]

PR #131, opened 2022-11-29
Miscellaneous specification fixes

mbolivar-nordic:
A few random fixes. In particular, this pull request:

Fixes: https://github.com/devicetree-org/lopper/issues/130
Fixes: https://github.com/devicetree-org/lopper/issues/128

See commit logs for details.


mbolivar-nordic:
@sstabellini the remaining issue I would like to fix this week is https://github.com/devicetree-org/lopper/issues/129, but I will need a bit more time today to think about your reference text and may have some follow up questions.

PR #153, opened 2023-02-08
lopper: Update r5 node names as per latest system device-tree changes

kedareswararao:
As per system device-tree specification each cpu cluster cpu node name should start from 0, update the code for the same.

Signed-off-by: Appana Durga Kedareswara rao <appana.durga.kedareswara.rao@amd.com>

### DIFF
--- a/lopper/assists/cfg_obj/modules/sdtinfo.py
+++ b/lopper/assists/cfg_obj/modules/sdtinfo.py
@@ -155,9 +155,9 @@ class SdtInfo:
                                     None
                 if node.name == "cpus-a53@0":
                     self.masters["psu_cortexa53_0"]["slaves"] = slaves
-                elif node.name == "cpus-r5@1":
+                elif node.name == "cpus-r5@0":
                     self.masters["psu_cortexr5_0"]["slaves"] = slaves
-                elif node.name == "cpus-r5@2":
+                elif node.name == "cpus-r5@1":
                     self.masters["psu_cortexr5_1"]["slaves"] = slaves
             except:
                 None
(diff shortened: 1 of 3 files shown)


mbolivar-nordic:
cc @zeddii I don't think the spec requires this:

- CPU cluster nodes had their reg properties and unit addresses removed in 2cfea13e5f54e9b334d987ef45b4ec27c5c8f316.
- CPU nodes obey the same semantics as /cpus/cpu* nodes in regular devicetree, and I don't know of anything in the DT spec that says they have to start at 0.  The requirements for those unit addresses are:

```
The value of *reg* is a ``<prop-encoded-array>``
that defines a unique CPU/thread id for the
CPU/threads represented by the CPU node.

If a CPU supports more than one thread (i.e.
multiple streams of execution) the *reg*
property is an array with 1 element per
thread. The *#address-cells* on the ``/cpus`` node
specifies how many cells each element of the
array takes. Software can determine the number
of threads by dividing the size of *reg* by
the parent node's *#address-cells*.

If a CPU/thread can be the target of an
external interrupt the *reg* property value
must be a unique CPU/thread id that is
addressable by the interrupt controller.

If a CPU/thread cannot be the target of an
external interrupt, then *reg* must be unique
and out of bounds of the range addressed by
the interrupt controller

If a CPU/thread's PIR (pending interrupt register)
is modifiable, a client
program should modify PIR to match the *reg*
property value. If PIR cannot be modified and
the PIR value is distinct from the interrupt
controller number space, the CPUs binding may
define a binding-specific representation of
PIR values if desired.
```

kedareswararao:
Thanks @mbolivar-nordic for the review will fix it in generic way...

PR #181, opened 2023-05-10
Spec fixes march 2023

mbolivar-nordic:
@zeddii This is a roll-up of the patches that I've sent to the list and which @sstabellini reviewed.

I left his reviewed-by off of `specification: domains: fix access property semantics` just to be conservative, although he did say he was OK with it with one suggested improvement, which I did make in this updated version of the patch.


zeddii:
merged!

PR #185, opened 2023-05-30
lopper: lops: lop-domain-linux-a53-prune: prune symbol node

kedareswararao:
After pruning the system device-tree for linux need to prune symbol node as well so that it will be in sync with the generated linux device-tree nodes, updated the logic for the same.

### DIFF
--- a/lopper/lops/lop-domain-linux-a53-prune.dts
+++ b/lopper/lops/lop-domain-linux-a53-prune.dts
@@ -56,6 +56,7 @@
                       code = "
                           # Check for domain node
                           from domain_access import update_mem_node
+                          from baremetalconfig_xlnx import get_label
                           domain_node = []
                           for n in __selected__:
                               if re.search('domains', n.name):
@@ -179,6 +180,18 @@
                                       prop_val.append(address_map[inx+2*na])
                               modify_val = update_mem_node(node1, prop_val)
                               node1['reg'].value = modify_val
+
+                              # Update symbol node referneces
+                              symbol_node = tree['/__symbols__']
+                              prop_list = list(symbol_node.__props__.keys())
+                              match_label_list = []
+                              for n in tree['/'].subnodes():
+                                  matched_label = get_label(tree, symbol_node, n)
+                                  if matched_label:
+                                      match_label_list.append(matched_label)
+                              for prop in prop_list:
+                                  if prop not in match_label_list:
+                                      tree['/__symbols__'].delete(prop)
                       ";
                 };
         };


lopper/lops/lop-domain-linux-a53-prune.dts
```diff
@@ -179,6 +180,18 @@
                                       prop_val.append(address_map[inx+2*na])
                               modify_val = update_mem_node(node1, prop_val)
                               node1['reg'].value = modify_val
+
+                              # Update symbol node referneces
+                              symbol_node = tree['/__symbols__']
+                              prop_dict = symbol_node.__props__.copy()
```
zeddii, line 186: Do you need the .copy() here ? The prop dict isn't being modified, so iterating the node properties should be enough, or am I missing something (and there is a lopper iterator that can be used to hide the internal node details).

I just want to be sure you aren't working around a lopper core bug, that I should be fixing.

lopper/lops/lop-domain-linux-a53-prune.dts
```diff
@@ -56,6 +56,7 @@
                       code = "
                           # Check for domain node
                           from domain_access import update_mem_node
+                          from baremetalconfig_xlnx import get_label
```
zeddii, line 59: We now have get_label() in two assists.  We can look at moving it into a lopper library to avoid the duplication.

Let me ask a question though (not related to this review, but for my information in consolidating the routine).

What is get_label() actually doing ? It is checking the node __props__ dictionary, which is an ordered dict of property names to LopperProp objects.

But yet the list comprehension / evaluation variables are 'label' and 'node_abs'. The node_abs[0], will get the property as a string (in the first chunk), so that would work.

I assume that the symbol node has properties that are name -> node path ? And that this is looking for a LopperProperty that points to the node being compared to.

The variables are a bit oddly named, but I hope I've got the intent correct.

lopper/lops/lop-domain-linux-a53-prune.dts
kedareswararao: Thanks @zeddii  for review.
Here my intention is to remove the node references in the symbol node for the nodes which are deleted by this lops file, to achieve this traversing the symbol node entries against each node getting the label list and traversing the symbol node entries and deleting them if there is no match against the label list.

if we won't copy  getting the error  Exception (OrderedDict mutated during iteration).

Will fix the code part without copy

lopper/lops/lop-domain-linux-a53-prune.dts
kedareswararao: Yes @zeddii  your understanding is correct, please let me know once the changes are available in the lopper library  will remove the get_label API and will update source code according to new implementation

lopper/lops/lop-domain-linux-a53-prune.dts
zeddii: I didn't see where that would happen on my first read through, but I can see now that the delete() being nested in the iteration of the prop_dict and the .delete() would cause that to happen.

I did of course run into this multiple times in the core of lopper, and to avoid needing a copy() I'd wrap the iterator in list()

That is a shallow copy, so it shouldn't be too expensive (and the list() isn't "free" either), but I tend to avoid copy() as inevitably I'll end up modifying the copy and not the original and wonder what broke!

I can merge this as-is, unless you want to try the list() technique to confirm if it does or doesn't work.

kedareswararao: Fixe with the list implementation and force pushed the commit, please review...

PR #191, opened 2023-06-19
lopper: Add initial PL Microblaze support

mubinsyed:
- Update assist file baremetallinker_xlnx to support mig7 memory in linker
  file generation
- Add lop-microblaze.dts lops file to generate compiler
  flags and library path for Microblaze processor.
  Microblaze compiler flags are tightly coupled with
  Microblaze processor configured in specific HW design.

  lop-microblaze.dts generates cflags.yaml based on CPU node
  properties, which contains compiler flags and standard library path.


zeddii:
Very nice, good use of the lops to collect and output. No concerns from me.

PR #196, opened 2023-07-05
assists/baremetal: Add handling for non homogeneous driver device trees V2

conallogriofaamd:
Drivers such as rfdc and the dfe drivers have nested structures which are not repeated.

SD_FEC has device tree paramaters that are an array.

This patch will allow handling for these kinds of drivers

### DIFF
--- a/lopper/assists/baremetalconfig_xlnx.py
+++ b/lopper/assists/baremetalconfig_xlnx.py
@@ -415,6 +415,7 @@ def xlnx_generate_config_struct(sdt, node, drvprop_list, plat, driver_proplist, 
                 plat.buf(' /* %s */' % prop)
 
 def xlnx_generate_prop(sdt, node, prop, drvprop_list, plat, pad, phandle_prop):
+    nosub = 0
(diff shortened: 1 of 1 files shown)


zeddii:
Can we get a quick v2 with a tweaked shortlog ?

This should be: assists/baremetal: add handling for non homogeneous driver device trees

github is also complaining about the signed-off-by, but I'm not seeing any issues with it myself.

conallogriofaamd:
> Can we get a quick v2 with a tweaked shortlog ?
> 
> This should be: assists/baremetal: add handling for non homogeneous driver device trees
> 
> github is also complaining about the signed-off-by, but I'm not seeing any issues with it myself.

Will do, Thanks!

conallogriofaamd:
> Can we get a quick v2 with a tweaked shortlog ?
> 
> This should be: assists/baremetal: add handling for non homogeneous driver device trees
> 
> github is also complaining about the signed-off-by, but I'm not seeing any issues with it myself.

Hi, Done. Thanks!

PR #203, opened 2023-07-14
lopper: assists: baremetalconfig_xlnx: Add support for variable size structs

conallogriofaamd:
Certain drivers have a number of nested structs, all with different numbers of members.
This patch adds support for this allowing for extra braces to be included if a split is defined in the child node.


zeddii:
The changes themselves look ok, but the 2nd non-merge commit is missing a Signed-off-by line.

I can rebase this before merging, or you can do that before a v2, either way works for me.

zeddii:
rebased, messages tweaked and merged to master.

PR #213, opened 2023-08-03
lopper: assists: baremetallinker_xlnx: linker updates

kedareswararao:
1. If the app template has linker_constraints i.e heap and stack size configuration generate the same in cmake meta-data.
2. Update the tcm memory node checks for all the processor instances

### DIFF
--- a/lopper/assists/baremetallinker_xlnx.py
+++ b/lopper/assists/baremetallinker_xlnx.py
@@ -216,11 +216,11 @@ def xlnx_generate_bm_linker(tgt_node, sdt, options):
     For cortexr5 processor in ZynqMP and Versal SOC TCM memory map is fixed
     update the memory section for the same.
     """
-    if machine == "psu_cortexr5_0":
+    if "psu_cortexr5" in machine:
         mem_sec += '\n\tpsu_r5_0_atcm_MEM_0 : ORIGIN = 0x0, LENGTH = 0x10000'
         mem_sec += '\n\tpsu_r5_0_btcm_MEM_0 : ORIGIN = 0x20000, LENGTH = 0x10000'
         mem_sec += '\n\tpsu_r5_tcm_ram_0_MEM_0 : ORIGIN = 0x0, LENGTH = 0x40000'
-    if machine == "psv_cortexr5_0":
+    if "psv_cortexr5" in machine:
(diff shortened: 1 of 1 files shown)


lopper/assists/baremetallinker_xlnx.py
```diff
@@ -177,8 +178,22 @@ def xlnx_generate_bm_linker(tgt_node, sdt, options):
...
+    appname = utils.get_base_name(app_path)
+    yaml_file = os.path.join(app_path, "data", f"{appname}.yaml")
+    stack_size = None
+    heap_size = None
+    if not utils.is_file(yaml_file):
+        print(f"{appname} doesn't have yaml file")
+    else:
+        schema = utils.load_yaml(yaml_file)
+        if schema.get("linker_constraints"):
+            if schema.get("linker_constraints")["stack"]:
```
onkarharsh, line 192: Hi @kedareswararao ,
I guess you intend to use:
    schema["linker_constraints"].get("stack")

And same for heap.

lopper/assists/baremetallinker_xlnx.py
```diff
@@ -282,4 +297,9 @@ def xlnx_generate_bm_linker(tgt_node, sdt, options):
         memip_list.insert(0, memip_list.pop(memip_list.index(has_ocm[0])))
     cfd.write("set(TOTAL_MEM_CONTROLLERS %s)\n" % to_cmakelist(memip_list))
     cfd.write(f'set(MEMORY_SECTION "MEMORY\n{{{mem_sec}\n}}")\n')
+    if stack_size:
+        cfd.write(f'set(STACK_SIZE {stack_size})\n')
```
onkarharsh, line 301: Where are we defining the default size of stack and heap? Is it possible to keep all of them at one place in here?

lopper/assists/baremetallinker_xlnx.py
kedareswararao: Thanks fixed it now...

lopper/assists/baremetallinker_xlnx.py
kedareswararao: default one we have kept in our linker_gen API...

kedareswararao: Fixed now with the latest commit added default stack and heap size based on the processor.

PR #226, opened 2023-09-12
lopper: assists: gen_domain_dts: Add assist for generating processor specific dts

kedareswararao:
When user provided processor name and system device-tree as input this assist produces as a dts file which contains nodes mapped to that processor instance.


zeddii:
If this ever needs to run in cooperation or in a pipeline with other assists, they pruning of the tree could be done in a copy of the passed SDT.

But that is something relatively easy to retrofit, so I have no big concerns.

PR #227, opened 2023-09-20
lopper: assists: Add support for pruned system device-tree

kedareswararao:
If the system devic-tree is already pruned no need to run the get_mapped_nodes() API update the assist logic for the same.


zeddii:
I like the approach. A good streamlining of the processing.

PR #229, opened 2023-09-21
openamp: xlnx: Enable multiple channels in YAML

bentheredonethat:
Enable multiple channels to be defined in YAML as either:
- 1-1
- 1-many

Also add option where zero-indexed arg can be provided to denote which channel the output header file is targeting. If none provided default to 0.

The option is as follows:
openamp [remote|host]  <channel-ID>

where this is channel ID is expected to be an int value.


zeddii:
Looks good to me. No concerns on my end.

PR #236, opened 2023-11-02
Openamp xlnx outputfile formatting

bentheredonethat:
update output file string values


bentheredonethat:
@zeddii  updating the text file in openamp-xlnx context here.

PR #238, opened 2023-11-07
openamp: xlnx: Add new module flags for role, host and remote

bentheredonethat:
Enable usage of module using explicit key value pairs as command line arguments

The arguments required now are:

--openamp_role=[host|remote]

Must be 'host' or 'remote'. This denotes the role which the output header file will target.

--openamp_host and --openamp_remote

These each must correspond to the relevant cores in a OpenAMP Channel Pair. On ZynqMP if a YAML has A53-0 as host and RPU0 as remote a valid command line set would be:

--openamp_role=host --openamp_host=a53_0 --openamp_remote=r5_0

### DIFF
--- a/lopper/assists/openamp_xlnx.py
+++ b/lopper/assists/openamp_xlnx.py
@@ -672,24 +672,6 @@ def xlnx_rpmsg_parse_generate_native_amba_node(tree):
 
     return amba_node
 
-def xlnx_rpmsg_parse_get_channel(options, len_remote_nodes):
-        channel = None
-        # specify Channel ID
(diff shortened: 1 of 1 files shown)


bentheredonethat:
@zeddii  this deprecates the CI test FYI. We will have to make sure that the roles are present in the base test.

Please feel free to ask if the new test arguments are unclear. The Lopper-output from the module should instruct you on valid inputs...

something along the lines of:

< other standard args....> \
 -- openamp --openamp_role=host --openamp_host=a53_0 --openamp_remote=r5_0

PR #245, opened 2023-11-30
lopper: Add support for Microblaze RISC-V baremetal BSP

mubinsyed:
This pull request does following,
1.  Adds lops file to generate Microblaze RISC-V compiler flags.
     Microblaze RISC-V compiler flags are tightly coupled with
     Microblaze RISC-V processor configured in specific HW design. lop-microblaze-riscv.dts generates cflags.yaml based on CPU 
     node properties, which contains compiler flags, part of linker flags dependent on HW and standard library path.
2. Fixes handling for lmb bram
    Existing logic assumes that LMB BRAM is always mapped at 0x0. It is incorrect, it can be mapped
    anywhere within processor address map.
    
    Also, initial 80 bytes of LMB BRAM would be used directly by vector sections in linker script, only
    in case of classic Microblaze processor. Hence, addition of 80 bytes offset to LMB BRAM should be done only in case
    of Microblaze case.
   This bug was found while testing one of the HW design targeting Microblaze RISC-V, where LMB BRAM is mapped at address 
   that is non zero.


mubinsyed:
Changes are merged through other pull request

PR #248, opened 2023-12-06
openamp: xlnx: Add IPI Helper code

bentheredonethat:
Remove dead code

Add IPI Info for each SOC as helper info


bentheredonethat:
also added versal-net helper info for ipi

PR #249, opened 2023-12-10
lopper:assists:baremetal_xparams Generate the Memory region names within NOC macros

onkarharsh:
Presently, the NOC address related macros come in a generic way. To inline with the existing macro names and make the existing Translation Table work, improvise the logic to add region names of all the possible regions for AXI_NOC and AXI_NOC2.


onkarharsh:

kedareswararao:
Changes looks fine to me

PR #259, opened 2024-01-05
Keep symbols

bentheredonethat:
This is for linux boot use cases 

CC @kedareswararao  - please review as well.


kedareswararao (commented):
Hi @bentheredonethat ,
What is the exact issue you are facing here we are not removing the entire symbol node, we are deleting the references to nonexistence nodes from the tree...

Regards,
Kedar.

PR #262, opened 2024-01-10
openamp: xlnx: Enable Linux generated Device Tree to work for OpenAMP…

bentheredonethat:
… use cases

Update pruning in gen_domain_dts so that:
- TCM banks remain. Need these for OpenAMP ELF Loading
- Remove extra IPI nodes. These conflict if other domains want to use IPI too.

Add lop file to invoke gen_domain_dts. Make sure it is present in lop-load.dts so that the assist gen_domain_dts can be invoked as a lop file. This enables user to control the order of assist invocations upon the input SDT.

Sample usage:
./lopper.py -f -v --enhanced --permissive     \
 -i <OpenAMP YAML>                            \
 -i lopper/lops//lop-load.dts                 \
 -i lopper/lops//lop-xlate-yaml.dts           \
 -i lopper/lops/lop-openamp-invoke.dts        \
 -i lopper/lops/lop-gen_domain_dts-invoke.dts \
 -i lopper/lops/lop-a72-imux.dts              \
 system-top.dts out.dtb


bentheredonethat:
CC  @kedareswararao @zeddii  for review

kedareswararao (approved):
Changes looks fine to me

zeddii:
The ordering solution that you came up with, is exactly the currently recommended one. Load the assist and then invoke it via lop (explicitly or via a trigger/matching node).

So no concerns from me either.

PR #264, opened 2024-01-11
assists:baremetalconfig: Avoid the error when there is no serial mapped to the processor

onkarharsh:
In cases where there are serial nodes in the SDT but none of them are mapped to the processor being targeted in the BSP, the assist should avoid generating STDIN and STDOUT macros. Presently, it is raising an exception that "'name' is not found for the serial_node". Fix the same by checking if the serial_node exists before the 'name' comparison happens with the other nodes.


onkarharsh:

kedareswararao (approved):

PR #267, opened 2024-01-25
lopper: assists: baremetal*: Add provision to pass float numbers from SDT as is

onkarharsh:
Device Tree doesn't allow a floating point number in it as is. This is leading to complexity in the baremetal use case as we have one to one mapping of data types between the device tree entries and the baremetal xparameters/config structures. Therefore, provide a flexibility to pass the floating number in device tree as a string. Add intelligence in the assist to deduce a floating number within string.


onkarharsh:

PR #270, opened 2024-02-01
yaml: Remove the unused import of "packaging" module

onkarharsh:
commit 738df8cc6819 (yaml: support ruamel versions > 0.17.x) added support to check the ruamel version before employing the yaml specific APIs. While doing so, packaging module was imported but it was never invoked. Moreover, if at all it had to be used, it should have been updated under the requirements in setup.py. This statement is leading to missing packaging module related errors as packaging doesnt come with all python versions by default.

Remove this unused import statement of packaging module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setuptools.setup(
     include_package_data=True,
     install_requires=[ "humanfriendly","configparser" ],
     extras_require={ "server": ["flask>=1.1.2","flask_restful>=0.3.8","pandas"],
-                     "yaml": ["pyaml","ruamel.yaml","anytree"],
+                     "yaml": ["pyaml","ruamel.yaml","anytree","packaging"],
                      "dt" : ["devicetree"],
                      "pcpp" : ["pcpp"],
                     },


zeddii:
No sorry. There are more pending changes that will use this.

setup.py was skipped for the requirement as a full release hasn't been rolled yet, and there was a push for quick resolution.

zeddii:
IF we want to update the setup.py to track the requirement that is acceptable, or I'll do it when I do the next lopper release shortly.

onkarharsh:
Hi @zeddii ,
In either case, we can't have an open source repo in a broken state. As the packaging module is not being used in its current state, it seems reasonable to have this change. Once we use the packaging module features in future, we can update the requirements as well as add the import statement.
Please weigh this in.

zeddii:
The repository is not broken. The only reports I've had of breakage from the change are from one set of users running Lopper. Other environments are showing passing status.

My objections are not directly related to the packaging module, it is about the floating state of some of the environments using Lopper.  The core can and will add new required modules, the environments that run Lopper need to have a way to update to match.

A newer ruenamel.yaml that was not in the Pipfile.lock is what triggered the need for version checks, and now we have issues with the module supporting extended version checks not being in the setup.py description. So either the module recommendations are being followed, or they aren't. We can't chose booth, or whatever suits at the moment.

I can push my commits that are fulling using the packaging module if required, but I'm trying to minimize the changes at the moment.

Or if I drop the module, it will break pending changes and will just have to address this again in a few weeks.

onkarharsh:
> The repository is not broken. The only reports I've had of breakage from the change are from one set of users running Lopper. Other environments are showing passing status.
> 
> My objections are not directly related to the packaging module, it is about the floating state of some of the environments using Lopper. The core can and will add new required modules, the environments that run Lopper need to have a way to update to match.

Agree @zeddii on this one. As and when we introduce the dependency on a new module, we should ideally update the requirements in setup.py at the same time. Our environment uses "pip install -e git+https://github.com/devicetree-org/lopper.git@master#egg=lopper[dt,pcpp,yaml]" to install lopper which is a valid pip use case. If we add the requirements at the same time we introduce a new module, no environment would break. IMHO, we should not restrict ourselves to update requirements only during the time we are about to host it over PYPI.

> 
> A newer ruenamel.yaml that was not in the Pipfile.lock is what triggered the need for version checks, and now we have issues with the module supporting extended version checks not being in the setup.py description. So either the module recommendations are being followed, or they aren't. We can't chose booth, or whatever suits at the moment.

TBH,I still didn't understand the scenario which led you to this version check. We could always limit the minimum required version of ruamel.yaml in setup.py and upgrade the lopper version instead of doing this check. 

> 
> I can push my commits that are fulling using the packaging module if required, but I'm trying to minimize the changes at the moment.
> 
> Or if I drop the module, it will break pending changes and will just have to address this again in a few weeks.

If you want me to send the update to setup.py, I am up for it. Please let me know your views.

zeddii:
> > The repository is not broken. The only reports I've had of breakage from the change are from one set of users running Lopper. Other environments are showing passing status.
> > My objections are not directly related to the packaging module, it is about the floating state of some of the environments using Lopper. The core can and will add new required modules, the environments that run Lopper need to have a way to update to match.
> 
> Agree @zeddii on this one. As and when we introduce the dependency on a new module, we should ideally update the requirements in setup.py at the same time. Our environment uses "pip install -e git+[https://github.com/devicetree-org/lopper.git@master#egg=lopper[dt,pcpp,yaml]](https://github.com/devicetree-org/lopper.git@master#egg=lopper%5Bdt,pcpp,yaml%5D)" to install lopper which is a valid pip use case. If we add the requirements at the same time we introduce a new module, no environment would break. IMHO, we should not restrict ourselves to update requirements only during the time we are about to host it over PYPI.

Hopefully this will be readable, the reply is showing as text only.

I definitely agree with this as well, I was in a rush to unbreak environments with newer ruamel (ones that weren't respecting the versions in the Pipfile.lock) and didn't have time to do any research or a survey to see if environments with an older ruamel could be updated. So to not break either configuration, the version check was added, as I can't force update the existing environments that may just be doing a 'git pull' to update Lopper.

Eventually version checks were going to be required, so I added them to cover all scenarios.

> 
> > A newer ruenamel.yaml that was not in the Pipfile.lock is what triggered the need for version checks, and now we have issues with the module supporting extended version checks not being in the setup.py description. So either the module recommendations are being followed, or they aren't. We can't chose booth, or whatever suits at the moment.
> 
> TBH,I still didn't understand the scenario which led you to this version check. We could always limit the minimum required version of ruamel.yaml in setup.py and upgrade the lopper version instead of doing this check.
> 
> > I can push my commits that are fulling using the packaging module if required, but I'm trying to minimize the changes at the moment.
> > Or if I drop the module, it will break pending changes and will just have to address this again in a few weeks.
> 
> If you want me to send the update to setup.py, I am up for it. Please let me know your views.

If we think that will unbreak the environment without the 'packaging' module, that would be ideal. But if they really can't install or update to it, then I agree we have no choice but to drop the import for now.

I just don't have access to the environment to test it myself.

I definitely can drop the new import, but I'm trying to make sure it is because we have no ability to get packaging installed in those environments, and I can understand better the issues for the next time I need to add something

Bruce

zeddii:
Definitely send the change, I can also push my more readable use of version checking using the package in that case.

Thanks for the investigation, that's what I needed to be convinced to complete the checks, as I didn't want to use it in more places if I needed to change tactics.

Either a new PR or just update this one, either works.

I can merge it right away.

And my apologies for breaking the setup, it is ironic that my efforts to keep all environments working broke some of them.

PR #280, opened 2024-02-15
lopper:assists:baremetal*: Correct the steps to construct a 64 bit value from 2 32-bit cells

onkarharsh:
There is often a use case to construct a 64-bit value from 2 32-bit cells in a device tree. Presently, ljust is being used to convert the lower cell value into a valid 32-bit number. Use of ljust for this is not taking the leading zeroes into account. It is appending 0s always at the end which means 0x927 and 0x9270000 entries of device tree becomes the same after processing. Fix this by replacing ljust with string formating to 32 bit values (:08x).


onkarharsh:
CC: @conallogriofaamd

kedareswararao (approved):
Changes looks fine to me.

PR #281, opened 2024-02-16
lopper: assists: Add special handling for xdma pcie

kedareswararao:
In the system device-tree generated xdma node there are lot of properties getting generated in the node, due to which linux dtb generation is failing due to size limiation. Fix this issue by reading the Linux driver yaml binding doc and keep only those properties in the node.

Note: This is workaround, We are planning to read the yamls from the Linux sources instead of copying it to assists folder it's TODO.


zeddii:
Thanks for the extra comments/context.

I was reviewing this on my mobile device and waited to get to keyboard to properly comment and say "We should really pass this as an input and/or use the linux version directly".

But you've already covered that in the comment, so I'm good with the change.

PR #282, opened 2024-02-19
lopper: assists: Enhance special property removal

conallogriofaamd:
1. Refactor to allow special handling for multiple IPs
2. Child node handling: If a child node does not have any of of the properties listed in the YAML (or a child with any of these properties), it will be removed from the device tree.
3. Add YAML for RFDC (second special handling)

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -176,7 +176,7 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
                             'psx_fpd_gpv']
 
     if linux_dt:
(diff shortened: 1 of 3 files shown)


kedareswararao:
> 1. Refactor to allow special handling for multiple IPs
> 2. Child node handling: If a child node does not have any of of the properties listed in the YAML (or a child with any of these properties), it will be removed from the device tree.
> 3. Add YAML for RFDC (second special handling)

@zeddii : Changes looks fine to me , @conallogriofaamd As i said in my earlier pull request this is workaround only we should get rid of these yamls once we pass the linux  kernel source directory as input to this yaml.

PR #283, opened 2024-02-20
lopper: assists: barmetal*: Fix race condition in the version driver handling

kedareswararao:
To get the driver name of the version driver we are using re.sub with v_* string there can be driver with v_* name example (v_vcresampler) for such drivers existing code is returning wron driver name fix this issue by using re.split().


kedareswararao:
@zeddii :Could you please review and merge this pull request if changes looks fine...

zeddii:
I somehow missed the pull request notification, thanks for the ping.

PR #286, opened 2024-02-21
lopper: assists: gen_domain_dts: Remove dupe TCM node

bentheredonethat:
As there are 2 of each TCM bank in the input SDT, prune the global TCM node for each register space.


bentheredonethat:
@kedareswararao  please review.

kedareswararao:
Changes looks fine to me...
@zeddii @bentheredonethat

bentheredonethat:
it seems like this is not  the right TCM node to delete in all cases... i'll have to revert this patch and send an update

PR #287, opened 2024-02-21
lopper: venv: add requirements.txt and venv usage in README

its-izhar:
Using python3's venv is very powerful and makes doing lopper development and usage easier.

More on it: https://docs.python.org/3/library/venv.html

Update the README to add usage instructions about this and a requirements.txt which makes installing the lopper python dependencies a breeze using pip.

Been using this flow for a while, it'd be a useful addition to lopper mainline so others can also benefit from it.

While at it, fix some minor typos in the README as well.


zeddii:
I'm going to sit on this one for a bit. I have another pypi release to do, and I'm hoping to update to some more modern packaging mechanisms than requirements.txt, etc

We already have the Pipfile and Pipfile.lock that are there for a similar purpose.

PR #299, opened 2024-03-11
lopper: assists: xlnx_overlay_dt: Handle phandle properties in a spec…

kedareswararao:
…ific way

Currenlty, in the assist, there is no way to know phandles that are specific to AMD/Xilinx. Generic phandles are defined in the phandle_possible_properties dictionary in the fdt.py file. Create a phandle_dict in the assist and update the phandle properties as needed.


zeddii:
I didn't realize that this assist was manually generating the DTS output (instead of using the lopper core routines).

So while I'm ok with this change, we should open an issue / feature request to convert this to build a lopper tree structure and then use the lopper core routines to output the DTS from the tree structure.

There may be some issues with symbol lookups, as this would be a secondary tree, but that's functionality that we can add to the core of lopper to be able to generate a separate overlay that can lookup symbols in the main dts.

PR #306, opened 2024-03-19
lopper: assists: gen_domain_dts: Add option to prune TCM nodes

bentheredonethat:
Add option that if present allows user to prune TCM DT nodes.

To remove TCM nodes, add third option on command line for this assist.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -71,6 +71,12 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
     except IndexError:
         pass
 
+    keep_tcms = None
+    try:
+        keep_tcms = options['args'][2]
+    except IndexError:
+        pass
+
     # Delete other CPU Cluster nodes
     cpunode_list = sdt.tree.nodes('/cpu.*@.*')
     clustercpu_nodes = []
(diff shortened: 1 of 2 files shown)


zeddii:
I'm ok with the change, but a couple of things did jump to mind.

As long as you are ok with not using a -- type option, I'm ok with it (i.e. humans aren't going to mess up the command line).

And a second thought, what the conditions where these should and shouldn't be pruned ? Is this something we could pick up programmatically ?

kedareswararao:
@bentheredonethat : instead of keeping tcm nodes default please delete it default and have them in case of openmap use case, please update the changes accordingly. 

CC: @onkarharsh

bentheredonethat:
> @bentheredonethat : instead of keeping tcm nodes default please delete it default and have them in case of openmap use case, please update the changes accordingly.
> 
> CC: @onkarharsh

what do you mean here?

what is a valid check? because this means that one assist has to be aware of the other effectively having been run.

bentheredonethat:
@kedareswararao  @onkarharsh  we need tcm nodes for openamp case
that is why i added switch
we can have the switch present by default

kedareswararao:
@bentheredonethat and @zeddii : latest changes looks fine to me.

Regards,
Kedar.

PR #319, opened 2024-04-15
Fix DDR Memory Allocation: Update Check for DDR Start and Size Adjustments

lakshmaiahm:
This pull request focuses on improving memory allocation by adjusting the start and size addresses specifically for DDR regions beginning at 0x0. Previously, adjustments were made for both DDR starting at 0x0 and 0x100000 (1MB), causing functionality issues during memory access operations that targeted the last 1MB of DDR.

With this update, the code will now only adjust start and size addresses for DDR regions that begin at 0x0. This change is aimed at resolving the problems related to memory access within the final 1MB of DDR space. By making this adjustment, we aim to enhance stability and performance in memory allocation and access operations.


zeddii:
I rebased and merged this to master and v0.2024.x without any merge-commits. If something is missing, send a follow up pull request.

PR #321, opened 2024-04-18
assists:baremetal_validate_comp_xlnx: Avoid the hw check when device_type property is missing in YAML

onkarharsh:
There is no need to do hardware check and report the error when the driver YAML is missing the device_type property. Trying this is leading to backward compatibility issues with previous version released sources.


onkarharsh:
Please merge it to v0.2024.x branch as well. TIA

PR #323, opened 2024-04-25
excluding TTC3 from peripheral test for R5 and R52 processors in versal, zynqmp and versal-net platforms

lakshmaiahm:
This patch modifies testapp to generate proper ttc IDs for peripheral
test to exclude ttc3 in perihperal test for psu_cortexr5 for zynqmp soc,
versal and versal-net as TTC3 is being used by sleep APIs in the BSP


zeddii:
merged to master and v0.2024.x

PR #327, opened 2024-05-20
assists: baremetal_xparameters: Generate Canonical definitions for IPI targets

Harini-T:
Current implementation doesnt generate canonical definitions for IPI targets.

Generate Canonical definitions of IPI Bitmask and Buffer Index for the available IPI targets.

@onkarharsh @kedareswararao can you please review ?

### DIFF
--- a/lopper/assists/baremetal_xparameters_xlnx.py
+++ b/lopper/assists/baremetal_xparameters_xlnx.py
@@ -202,7 +202,7 @@ def xlnx_generate_xparams(tgt_node, sdt, options):
                                 else:
                                     if node.propval('xlnx,ip-name') != ['']:
                                         ip_name = node.propval('xlnx,ip-name', list)[0]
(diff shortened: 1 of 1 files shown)


kedareswararao:
Changes looks fine to me

onkarharsh (changes requested):
Hi Harini, Please send v2 as discussed offline.
1. Try to avoid using ip-name in here, instead use drv name. We may miss updating ip names here with new platforms.
2. Try reducing lines of code by using a new if condition within existing child,required condition. Repetition of loop should be fine.

Harini-T:
Hi @onkarharsh @kedareswararao 
I have sent v2 in the same pull request
Can you please review it?

onkarharsh:
Change looks fine now

kedareswararao (approved):
Changes looks fine.

PR #331, opened 2024-05-30
openamp: Add Validation for carveouts and linker

bentheredonethat:
This patch does 2 things:
1. Validate RPMsg carveouts to ensure they are present for host and remote OpenAMP Domains
2. Once carveouts are validated, enable the DDR OpenAMP carveout for ELF Load to propagate to generated linker. Here this means that if linker module has 'openamp' set then attempt to get relevant OpenAMP DDR node's base and length which will be used in linker-metadata.


bentheredonethat:
@onkarharsh @kedareswararao  please review
@zeddii  please wait for above review

kedareswararao (approved):
Changes looks fine to me

PR #336, opened 2024-06-26
assists: linker_xlnx: Fix memory assignment to avoid build failure

lakshmaiahm:
This commit updates the logic to only include RAM memories ('ps7_ram_0' and 'ps7_ram_1') by modifying the condition to match '_ram' in the memory names.


lakshmaiahm:
need to do some changes in patch

PR #337, opened 2024-06-28
assists: linker_xlnx: Fix memory assignment to avoid build failure

lakshmaiahm:
This commit addresses a build failure in the `memory_test` application on the ZC706 board with the preset DMA design. The issue arises from the incorrect inclusion of non-RAM memories in the linker script, which causes insufficient memory during the build process.

To resolve this, the logic is updated to only include RAM memories (`ps7_ram_0` and `ps7_ram_1`) by modifying the condition to match '_ram' in the memory names. This ensures that only the appropriate memory regions are assigned, preventing the build failure.

Additionally, support is added for the memory test application for MicroBlaze designs.


onkarharsh:
Reviewed

PR #341, opened 2024-07-17
assists: baremetallinker: Update the logic to find phandle in addressmap

sivadur:
The current logic to find phandle in address map is actually checking against all fields in addressmap(address, size, phandle) which is incorrect. Fix this by checking it only against the phandle in addressmap.


kedareswararao (approved):

zeddii:
This looks ok to me.

But also note that there are LopperTree routines that can simplify (or at least unify) some of the logic in the assist.

i.e. see phandle_map() and how it is used in other places in the tree.

PR #342, opened 2024-07-18
assists: baremetallinker: Fix ddr_default logic for memory tests

sivadur:
For memory tests, the default_ddr is always getting assigned to zeroth bram even when we have multiple brams in the design. Ideally, it is expected to be assigned to bram with higher size, update the logic accrodingly.


sivadur:
I would be sending an updated patch for it

PR #345, opened 2024-07-21
fdt: generate __symbols__ when compile

chuangzhu:
It seems to me that many lops under `lopper/lops/` requires a `__symbols__` node to run. However, the `dtc` on my machine only generates the `__symbols__` node when `-@` is passed


zeddii:
This isn't something we want on by default as it causes issues in mult-platform support. Lops that require __symbols__ are passing this on the command line.

Perhaps at sometime in the future we can switch the default, but that time isn't now.

Theoretically it could be hidden behind a lopper flag/switch, so I'll consider that idea a bit more.

PR #346, opened 2024-07-21
lopper: don't check the existence of cpp when LOPPER_CPP is set

chuangzhu:
When LOPPER_CPP is set to other C preprocessors, e.g. aarch64-none-elf-cpp in a cross compilation environment, lopper still checks the existence of `cpp` and refuses to run if not exists.

Also fixed lopper passing an unrecognized argument `--passthru-comments` when LOPPER_CPP is set to C processors other than `pcpp`.

### DIFF
--- a/lopper/__init__.py
+++ b/lopper/__init__.py
@@ -122,9 +122,9 @@ class LopperSDT:
 
         # check for required support applications
         if libfdt:
-            support_bins = [ "dtc", "cpp" ]
+            support_bins = [ os.environ.get("LOPPER_DTC", "dtc"), os.environ.get("LOPPER_CPP", "cpp") ]
         else:
-            support_bins = [ "cpp" ]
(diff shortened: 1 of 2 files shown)


zeddii:
In the case of cpp, it is expected that lopper uses the host cpp, so it shouldn't have the cross prefix. 

How are you running lopper such that you need the cross prefix versus a host or native cpp ?

zeddii:
I should clarify that the 2nd chunk about pcpp is definitely needed (as is likely the first). I'm just trying to understand how the cross tools are coming into play, so I can consider if there's more lurking problems.

zeddii:
Ahah. Thanks!

That helps me understand that we aren't going out of our way to support a really strange runtime that will only lead to more issues once preprocessing is complete.

Although I still do insist that the host cpp is the right thing to use, you are absolutely correct that if the ability to change cpp is advertised through the LOPPER_CPP variable, there's no harm in checking it first, since the fallback of 'cpp' is still there.

PR #356, opened 2024-08-12
lopper: assists: Initial support for Versal Gen2 pl configuration

NavaKishoreManne:
It adds initial support for Versal Gen2 pl configuration use cases.


NavaKishoreManne:
Please review @kedareswararao

PR #357, opened 2024-08-14
assists: gen_domain_dts: Hardcode DDR address for Zynq to always start from 0

onkarharsh:
Yocto project expects zynq DDR base addresses to start from 0x0 for linux to boot. This has been a legacy expectation which is logically flawed but working. Add the temporary workaround to unblock them until the QEMU issue and the yocto integration issues are resolved.


kedareswararao (approved):
Changes looks fine to me

PR #366, opened 2024-09-03
lopper/tree: add get() for LopperTree class

onkarharsh:
Add a new function get() for LopperTree class to mock the builtin dictionary get() function. This function helps in avoiding the KeyError while accessing any node from the tree. This feature already exists for LopperNode (propval()) and LopperProp (__getitem__), but it is missing for LopperTree.

### DIFF
--- a/lopper/tree.py
+++ b/lopper/tree.py
@@ -3291,6 +3291,33 @@ class LopperTree:
         # not currently supported
         pass
 
+    def get(self, key, default = None):
+        """mocks the builtin dictionary get() function for Lopper tree object:
+
+            <Lopper Tree Object>.get(<node path>)
+
+        Returns the same output as that of __getitem__ described above when the passed
+        key matches with some node inside the tree object.
+
+        Returns the default value when the node is not valid for the tree and avoids
+        the standard KeyError exception.
+
+        Args:
+            key: string, int or LopperNode
+            default: any valid data type
+
+        Returns:
+           LopperNode object if the key is found in the object
+           default value when key is absent.
+
+        """
+
+        try:
+            return self.__getitem__(key)
+        except KeyError:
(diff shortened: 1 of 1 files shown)


zeddii:
I've considered adding this in the past as well, but the throwing of the exception was an original design point (but admittedly, it is no longer as important) .. but it of course makes all the calling code required the try/except blocks.

I don't object to the new interface, but I think there are some different options that would make it more in-line with the rest of the interfaces (not that they are currently completely consistent, but that  is always a goal).

The intent of the nodes() call was to do what you are attempting with the function in this commit, allow a lookup by path to be safe with respect to exception processing. In that you call it with a tighter regex to safely get  a single, specific match returned (and as you  would have seen __getitem__ uses just that technique internally.

My alternate proposals  are:
  - add a  "strict" parameter to nodes() such that the regex could be made more strict and the exception is caught when it is True
  - if we did add a new API, I'd call it nodeval() to match propval()

onkarharsh:
Thanks @zeddii . I am fine with reusing of nodes() if we can add a silent_discard kind of option to it.
nodes() try block uses: matches = [self.__nodes__[nodename]] which throws an exception when nodename is not found. Exception block should check for silent_discard option first and silently return None if that is enabled. Otherwise, it can continue checking the regex.

Let me know your view on this. Could this be covered as part of your tweak of lop's "modify" feature, the one that we are discussing over mail?

zeddii:
Can you check my prototype commit on the master-next branch ? It adds strict matching to nodes(). We can tweak further if it doesn't fix your issue.

onkarharsh:
Hi @zeddii , the prototype works fine for my use case. Closing this Pull Request. Please merge your fix. Thanks.

PR #367, opened 2024-09-04
lops:lop-cpu-oslist Add a new lops file to get ai_engine as cpu

Dhriti13:
Add a new lops file that can generate ai_engine as a cpu. Update the earlier cpulist.yaml with supported os data as well.

### DIFF
--- a/lopper/lops/lop-cpu-oslist.dts
+++ b/lopper/lops/lop-cpu-oslist.dts
@@ -1,0 +1,56 @@
+/*
+ * Copyright (C) 2024 Advanced Micro Devices, Inc.  All rights reserved.
+ *
+ * Author:
+ *     Dhriti Sree Rompicharla <DhritiSree.Rompicharla@amd.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+*/
+
+/dts-v1/;
+
+/ {
+        compatible = "system-device-tree-v1,lop";
+        lops {
+                compatible = "system-device-tree-v1,lop";
+                lop_0 {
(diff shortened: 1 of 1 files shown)


Dhriti13:
@kedareswararao and @sivadur, please review.
I have tested it against MB, ZynqMP, Zynq, Versal, Versal+MB, Versal with ai_engine, VersalNet designs.

lopper/lops/lop-cpu-oslist.dts
```diff
@@ -0,0 +1,56 @@
...
+                      lop_0_1 {
+                              compatible = "system-device-tree-v1,lop,code-v1";
+                              code = "
+                                     import yaml
+                                     yaml_dict = {}
+                                     for node in __selected__:
+                                        cpu_ip_name =  node.propval('xlnx,ip-name')[0]
+                                        yaml_dict[node.label] = {'ip_name' : '','supported_os' : []}
+                                        yaml_dict[node.label]['ip_name'] = cpu_ip_name
+                                        if 'cortex' in cpu_ip_name or 'microblaze' in cpu_ip_name:
```
sivadur, line 29: I think this will add freertos also as supported os for Microblaze-V which is not true. MB-V contains ip-name as "microblaze_riscv".  Can you please cross check

Dhriti13: Hi DP, Thanks for the review. Updated the logic and force pushed. 
Please check it once.

kedareswararao (approved):
Changes looks fine to me

PR #368, opened 2024-09-05
lopper_sanity: add openamp sanity test

bentheredonethat:


zeddii:
We should do one minor tweak.

When --all is passed, enable the openamp tests.

It is in the same case statement where you added the new enable flag.

bentheredonethat:
@zeddii  tweaked

PR #373, opened 2024-09-14
assists: baremetalinker: Update microblaze-v default address

sivadur:
Microblaze-V has configurable reset vector and hence update the default linker address to the memory section to which the reset vector is pointing to.

### DIFF
--- a/lopper/assists/baremetallinker_xlnx.py
+++ b/lopper/assists/baremetallinker_xlnx.py
@@ -320,6 +320,16 @@ def xlnx_generate_bm_linker(tgt_node, sdt, options):
     has_ocm = None
     has_ram = None
     has_bram = None
+
+    if "microblaze_riscv" in cpu_ip_name:
+        mbv_reset_addr = match_cpunode.propval('xlnx,base-vectors')[0] << 32 | match_cpunode.propval('xlnx,base-vectors')[1]
+        for key, value in mem_ranges.items():
+            end = value[0] + value[1]
+            if mbv_reset_addr == value[0]:
+                default_ddr = key
+                memtest_config = False
+                break
+
     ## For memory tests configuration default memory should be ocm if available
     if memtest_config:
         has_ocm = [x for x in mem_ranges.keys() if "ocm" in x]


lopper/assists/baremetallinker_xlnx.py
```diff
@@ -320,6 +320,16 @@ def xlnx_generate_bm_linker(tgt_node, sdt, options):
     has_ocm = None
     has_ram = None
     has_bram = None
+
+    if "microblaze_riscv" in cpu_ip_name:
+        mbv_reset_addr = match_cpunode.propval('xlnx,base-vectors')[0] << 32 | match_cpunode.propval('xlnx,base-vectors')[1]
+        for key, value in mem_ranges.items():
+            end = value[0] + value[1]
+            if mbv_reset_addr == value[0]:
+                default_ddr = key
```
kedareswararao, line 329: why are we changing the memtest_config to false here? to pick the highest memory for linker generation?

kedareswararao (commented):
memtest_config = Flase will make the assist to pick the highest size available memory to be picked by linker and also https://github.com/devicetree-org/lopper/blob/master/lopper/assists/baremetallinker_xlnx.py#L366 cmake meta-data contians data about all the memories which may cause runtime failure

lopper/assists/baremetallinker_xlnx.py
sivadur: @kedareswararao If i didnt make memtest_config as False, it will again make either ocm, ram, bram as default memory in linker which i would want to avoid and always make the memory to which reset vector is pointing to as default memory for MB-V

  if memtest_config:
        has_ocm = [x for x in mem_ranges.keys() if "ocm" in x]
        has_ram = [x for x in mem_ranges.keys() if "_ram" in x]
        has_bram = [x for x in mem_ranges.keys() if "_bram" in x]
        if has_ocm:
            default_ddr = has_ocm[0]
        elif has_ram:
            default_ddr = has_ram[0]
        elif has_bram:
            if len(has_bram) > 1:
                size = 0
                for key, value in mem_ranges.items():
                    if "_bram" in key:
                        if size < value[1]:
                            size = value[1]
                            default_ddr = key
            else:
                default_ddr = has_bram[0]

kedareswararao: @sivadur Okay as long as the memory node for which we are generating shouldn't be part of memory test application memory_config_g.c file if it's taken care i am fine with the changes, please let me know if it's taken care in the generated _g.c file will approve the changes.

zeddii:
FYI: I was still waiting to here if everyone is ok with this change.

sivadur:
closed this for now. If needed i will create a new pull request

PR #375, opened 2024-09-19
lopper: assists: baremetal_gentestapp_xlnx: Add support for dependency files

VenkateshOdela:
Add logic to read dependency files.
Add logic to find the valid example based on axistream-connected value. 
Add logic to detect and handle 'xlnx,eth-dma' and 'xlnx,eth-mcdma' nodes.

This patch ensures all dependency files are included and the correct example is selected for Axi-Ethernet in the peripheral test.


lopper/assists/baremetal_gentestapp_xlnx.py
```diff
@@ -248,6 +286,8 @@ def xlnx_generate_testapp(tgt_node, sdt, options):
         for app in testapp:
             if 'SelfTest' in app or 'selftest' in app:
                 status_assignment = f"status = {app}({xpar_def});"
+            elif node == "axi_ethernet":
```
kedareswararao, line 289: node name will vary design to design here you should depend on driver name instead of node name

lopper/assists/baremetal_gentestapp_xlnx.py
```diff
@@ -248,6 +286,8 @@ def xlnx_generate_testapp(tgt_node, sdt, options):
...
+                status_assignment = f"status = {app}(&{node}, &{dma_label}, {xpar_def});"
```
kedareswararao, line 290: what if dma_label is not present should we assign some dummy entry or through proper error?

lopper/assists/baremetal_gentestapp_xlnx.py
```diff
@@ -222,6 +249,17 @@ def xlnx_generate_testapp(tgt_node, sdt, options):
...
+                                    dep_file_dst_path = os.path.join(sdt.outdir, dep_file)
+                                    if utils.is_file(dep_file_src_path):
+                                       periph_file_list += [dep_file_dst_path]
+                                       utils.copy_file(dep_file_src_path, dep_file_dst_path)
+                                       with open(dep_file_dst_path, 'r+') as fd:
+                                           content = fd.readlines()
+                                           content.insert(0, "#define TESTAPP_GEN\n")
+                                           fd.seek(0, 0)
+                                           fd.writelines(content)
                                 dec.append(testapp_schema[app]['declaration'])
```
kedareswararao, line 263: This code looks fine

lopper/assists/baremetal_gentestapp_xlnx.py
```diff
@@ -128,6 +130,17 @@ def xlnx_generate_testapp(tgt_node, sdt, options):
...
+           for node in node_list:
+               if "xlnx,eth-dma" in node["compatible"].value:
+                   dma_node_list.append(node)
+                   dma_label = get_label(sdt, symbol_node, node)
+                   testapp_name.update({dma_label: 'XAxiDma'})
+               if "xlnx,eth-mcdma" in node["compatible"].value:
+                   dma_node_list.append(node)
+                   dma_label = get_label(sdt, symbol_node, node)
+                   testapp_name.update({dma_label: 'XMcdma'})
+
```
kedareswararao, line 143: We have a axi_fifo use case as well AFAIK are we not supporting it in the driver examples?

VenkateshOdela: Currently we don't have support for axi_fifo.

kedareswararao (approved):
Latest Changes looks fine to me

PR #376, opened 2024-09-20
assists:baremetallinker: Avoid setting axi_emc as the default memory in linker sections

onkarharsh:
axi_emc is a softIP used to interface with external memory devices. The block provides memory controller functionality primarily for parallel NOR Flash and linear flash devices. ELFs can't be directly downloaded over axi_emc using "dow" and a separate flash utility is needed. Therefore, avoid setting axi_emc as the default memory in all the linker sections. This is a makeshift solution until all the data and code sections are properly divided among the available memory IPs.

### DIFF
--- a/lopper/assists/baremetallinker_xlnx.py
+++ b/lopper/assists/baremetallinker_xlnx.py
@@ -166,11 +166,22 @@ def get_memranges(tgt_node, sdt, options):
 
     return mem_ranges
 
+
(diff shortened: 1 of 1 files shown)


onkarharsh:
@kedareswararao
@sivadur : your other PR may need a rebase before v2.

kedareswararao (approved):

PR #393, opened 2024-10-21
lopper: assists: gen_domain_dts: Preserve TCM nodes with latest checks

bentheredonethat:
After commit 0551671 the TCM nodes were still deleted. Add check to preserve TCM nodes.


bentheredonethat:
CC @onkarharsh

PR #413, opened 2024-11-26
lopper: assists: common_utils: Added logger setup based on verbose

sathishkumar-amd:
lopper: assists: baremetalconfig_xlnx: For scan_reg_size added exception to handle the Index error and added logger warnings
lopper: assists: baremetal_xparameters_xlnx: Added logger messages and removed duplicate code
lopper: assists: gen_domain_dts: Removed duplicate code

### DIFF
--- a/lopper/assists/baremetal_xparameters_xlnx.py
+++ b/lopper/assists/baremetal_xparameters_xlnx.py
@@ -21,9 +21,9 @@ import common_utils as utils
 import baremetalconfig_xlnx as bm_config
 from baremetaldrvlist_xlnx import xlnx_generate_bm_drvlist
 from baremetallinker_xlnx import get_memranges
+from lopper.log import _init, _warning, _info, _error, _debug, _level, __logger__
 
-
-logger = logging.getLogger(__name__)
+_init(__name__)
 
 def is_compat( node, compat_string_to_test ):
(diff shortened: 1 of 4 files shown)


sathishkumar-amd:
> The lopper codebase is (slowly, it is only being converted as other work is being done in the areas) to a consistent logging format and set of controls.
> 
> An example for an assist is isospec:
> 
> lopper/assists/isospec.py:from lopper.log import _init, _warning, _info, _error, _debug lopper/assists/isospec.py: lopper.log._init( **name** ) lopper/assists/isospec.py: lopper.log._level( logging.INFO, **name** ) lopper/assists/isospec.py: lopper.log._level( logging.DEBUG, **name** ) lopper/assists/isospec.py: lopper.log._level( logging.INFO, **name** )
> 
> Rather than using the logging/logger directly, these should be used whenever possible.
> 
> If there's a bug or other issue preventing their use, then definitely report it and I'll get it fixed.

Hi,

As suggested changed the logger logic and used from lopper/log

zeddii:
We could squash the commits into one, but I appreciate the quick turn around on the switch to lopper logging update. Having the history of the logger -> lopper logging also isn't a bad thing, since it can serve as a reference.

No need to squash this, I'll merge it as-is.

PR #421, opened 2024-12-12
lopper: assists: gen_domain_dts: Add Ignore IP entries for Versal Gen2 linux domain

onkarharsh:
IP Names have been updated from Versal Gen2 family onwards. There will no longer be any psu/psv/psx suffixes in the IP Names for Versal Gen2 family. Add Versal Gen2 specific IP Name entries in the ignore list to avoid un-needed IPs in linux domain and keep the linux domain device tree size in check.


onkarharsh:
Hi @kedareswararao ,
Please check the IP list once.

kedareswararao (approved):
Changes looks fine to me

PR #422, opened 2024-12-16
lopper : Fixed path issues

sathishkumar-amd:
The PurePosixPath was causing problems in the Windows flow since it is a supported method for Linux.
Change the logic from "PurePosixPath( output_filename ).is_absolute()" to "os.path.isabs( output_filename )" to support both Windows and Linux flows.

### DIFF
--- a/lopper/__init__.py
+++ b/lopper/__init__.py
@@ -12,7 +12,6 @@ import os
 import re
 import shutil
 from pathlib import Path
-from pathlib import PurePosixPath
 from io import StringIO
 import contextlib
 from importlib.machinery import SourceFileLoader
@@ -567,7 +566,7 @@ class LopperSDT:
             tree_to_write = self.tree
 
         if re.search( "\.dtb$", output_filename ):
-            if self.outdir and not PurePosixPath( output_filename ).is_absolute():
+            if self.outdir and not os.path.isabs( output_filename ):
                 output_filename = self.outdir + "/" + output_filename
 
             if self.use_libfdt:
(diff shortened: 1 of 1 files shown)


zeddii:
I need a much better description of the problem on windows. Is it slashes ? Is it something else ?

PurePosixPath is the more modern way to handle paths, and I would rather not go back to the os.* methods. I explicitly switched away from them early in Lopper's development.

If I understand more about what is going wrong, I can see about both sanity tests and alternatives (even though we may just end up right back here with this commit).

sathishkumar-amd:
> I need a much better description of the problem on windows. Is it slashes ? Is it something else ?
> 
> PurePosixPath is the more modern way to handle paths, and I would rather not go back to the os.* methods. I explicitly switched away from them early in Lopper's development.
> 
> If I understand more about what is going wrong, I can see about both sanity tests and alternatives (even though we may just end up right back here with this commit).

In Windows, an absolute path in the native file system typically begins with a drive letter (e.g., C:\, D:\, etc.). PurePosixPath is designed to emulate POSIX-style path behavior. In POSIX systems, an absolute path is defined by the presence of a leading "/".

Here are some examples in windows:

```
>>> if PurePosixPath("C:/test/new/").is_absolute():
...     print(True)
...
>>> if PurePosixPath("/C:/test/new/").is_absolute():
...     print(True)
...
True
```

In the first example, "C:/test/new/" is not considered an absolute path in the POSIX context. However, in the second example, "/C:/test/new/" is treated as an absolute path due to the leading slash.

zeddii:
From my testing, the simplest thing to do is just change my use of PurePosixPath to Path(), that abstraction will return true for  <drive letter>:/<path> on Windows, and will also return true for /foo/bar on Linux. Since pathlib does the switching from WindowsPath to PosixPath objects behind the scenes.

PR #423, opened 2024-12-16
lopper: assists: Optimized metadeta generation logic

lakshmaiahm:
Updated logic to remove unnecessary timer instances in metadata list


lakshmaiahm:
Updated logic to remove unnecessary timer instances in metadata list

PR #427, opened 2024-12-20
openamp: xlnx: Enable gen_domain to invoke openamp processing

bentheredonethat:
For gen_domain_dts.py plugin, if the openamp domains are present then have this processing occur at the start.

Additionally if the openamp processing ocurred then keep TCM nodes.

For OpenAMP to enable this do the following:
1. If OpenAMP processing occurs then at the end remove the OpenAMP nodes
2. Only output the OpenAMP Configuration header for non-Linux targets
3. Add some OpenAMP lookup tables to aid the gen_domain_dts.py in OpenAMP info lookup

This way the following use cases are still preserved
1. Non-Linux OpenAMP runs
2. If a user keeps the OpenAMP plugin args in a Linux domain Lopper run

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -22,6 +22,8 @@ from baremetalconfig_xlnx import compat_list, get_cpu_node, get_mapped_nodes, ge
 from common_utils import to_cmakelist
 import common_utils as utils
 from domain_access import update_mem_node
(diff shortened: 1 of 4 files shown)


bentheredonethat:
@zeddii  please wait to merge

@kedareswararao  @onkarharsh please review first. 
thanks

kedareswararao (approved):

bentheredonethat:
@zeddii please proceed

PR #432, opened 2024-12-31
lopper_sanity: xlnx: Add sanity test for simple gen_domain case

bentheredonethat:
@onkarharsh  @kedareswararao  please review before  merge

CC @zeddii

### DIFF
--- a/lopper_sanity.py
+++ b/lopper_sanity.py
@@ -2120,6 +2120,44 @@ def openamp_sanity_test_generic( sdt, overlay, output_sdt, target_soc, test_str,
     else:
         test_failed(test_str)
 
+def xlnx_gen_domain_sanity_test( verbose ):
+    ws_area = os.getcwd()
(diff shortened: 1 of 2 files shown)


lopper_sanity.py
```diff
@@ -2468,6 +2506,7 @@ def usage():
     print('  -t, --tree          run lopper tree tests' )
     print('  -l, --lops          run lop tests' )
     print('  -o, --openamp       run openamp tests')
+    print('  -x, --xlnx          run xilinx-amd sanity tests')
```
zeddii, line 2509: We should use a different / more generic name for this. If we start adding each user's sanity tests via a name, it will both the a maintenance issue and will be too specific.

Can we just call this "generate-domain"

Also, it needs to be run when --all is passed, which is currently not the case.

bentheredonethat:
@zeddii  updated

zeddii:
I closed the other dependent pull request, can you reactivate that pull rebased onto the current master ?

PR #433, opened 2025-01-01
Openamp xlnx baremetallinker handling

bentheredonethat:
first commit is present in other PR


zeddii:
The first commit needs a bit of tweaking based on the review I did there. But this commit looks fine. I'd suggest a new PR or an updated one with both and I can merge from there.

PR #434, opened 2025-01-02
xlnx_overlay_dt.py: Fix interrupt-parent assignment logic in overlay …

NavaKishoreManne:
…generation

Refined the interrupt-parent handling logic to overcome the limitations of the previous implementation, which was restricted to GIC. The updated logic now ensures compatibility with both GIC and axi-intc use cases.

### DIFF
--- a/lopper/assists/xlnx_overlay_dt.py
+++ b/lopper/assists/xlnx_overlay_dt.py
@@ -286,11 +286,11 @@ def xlnx_generate_overlay_dt(tgt_node, sdt, options):
                             plat.buf('%s' % phandle_prop)
                         elif re.match("clocks =", str(p)):
                             plat.buf('%s' % node['clocks'])
-                        elif p.name == "interrupt-parent":
-                            if gic_node and imux_node:
-                                if p.value[0] == imux_node.phandle:
-                                    plat.buf('%s' % str(p).replace('imux', 'gic'))
                         else:
+                            if p.name == "interrupt-parent":
+                                if gic_node and imux_node:
+                                    if p.value[0] == imux_node.phandle:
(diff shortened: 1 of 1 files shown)


kedareswararao (approved):

kedareswararao (approved):

lopper/base.py
```diff
@@ -474,7 +474,7 @@ def phandle_possible_properties(cls):
                     "address-map" : [ '#ranges-address-cells phandle #ranges-address-cells #ranges-size-cells', 0 ],
                     "secure-address-map" : [ '#ranges-address-cells phandle ^:#address-cells #ranges-size-cells', 0 ],
                     "interrupt-parent" : [ 'phandle', 0 ],
-                    "interrupts-extended" : [ 'phandle field' ],
+                    "interrupts-extended" : [ 'phandle field field' ],
```
zeddii, line 477: Do all occurrences of interrupts-extended follow this format ?

The link to the specification/binding implies that, but since it was currently only described as having a single field, were there no inputs that tested this, or were the existing inputs being incorrectly processed ? (and hence why this was noticed).

FYI. In 2025, I'll switch to reading bindings alongside the lopper specific description and this will be easier to catch.

NavaKishoreManne: The "interrupt-extended" property appears to have remained untested until now. According to the binding documentation, it expects two fields. After applying this fix, I verified the axi-intc use case, and it is functioning as expected.

kedareswararao: @zeddii looks like no inputs tested this property (interrupts-extended) as per binding docs this fix is proper, please merge it to avoid issues with the overlay use cases.

PR #445, opened 2025-01-22
lopper: assists: gen_domain_dts: Update memory node handling to consi…

kedareswararao:
…der device_type field

The current memory logic retrieves memory nodes by reading nodes named "memory@," which inadvertently leads to the deletion of reserved memory nodes with similar names. To prevent this, traverse only the memory nodes where device_type is explicitly set to "memory."


onkarharsh:
Looks Good to me.

PR #451, opened 2025-01-23
Openamp xlnx udates jan23

bentheredonethat:


bentheredonethat:
@zeddii  as part of this i updated subsystem assist to add a lockstep flag in YAML and update OpenAMP YAML parsing of ELFLOAD parsing

zeddii:
Thanks for the heads up. The subsystem tweak looks good to me.

PR #456, opened 2025-02-03
openamp: xlnx: Enable output of header given OpenAMP DT

bentheredonethat:


bentheredonethat:
@zeddii  let me add one more commit on top. thanks

PR #460, opened 2025-02-06
lopper: assists: baremetallinker_xlnx: Eliminated unnecessary code bl…

sathishkumar-amd:
…ocks that were no longer needed.

- Separated the 'xlnx_generate_bm_linker' code into the following functions to improve code flow:
   - **generate_mem_config**: Generates the memory configuration.
   - **get_ddr_address**: Retrieves the default DDR address value.
   - **generate_linker_script**: Contains the generic code for the linker script.
- Add the full names of the memory sections to the linker script to ensure proper alignment with Vitis Classic. 
- lopper: assists: Integrated label names within the 'get_memranges' function, which returns both the memory ranges and label names as part of adding the full names of the memory sections.

### DIFF
--- a/lopper/assists/baremetal_bspconfig_xlnx.py
+++ b/lopper/assists/baremetal_bspconfig_xlnx.py
@@ -37,7 +37,7 @@ def xlnx_generate_bm_bspconfig(tgt_node, sdt, options):
     if options.get('outdir', {}):
         sdt.outdir = options['outdir']
    
-    mem_ranges, lable_names = get_memranges(tgt_node, sdt, options)
+    mem_ranges, _ = get_memranges(tgt_node, sdt, options)
     if not mem_ranges:
         return
     # Generate Memconfig cmake meta-data file.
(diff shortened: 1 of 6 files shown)


kedareswararao (changes requested):
get_memranges API being used in the petalinuxconfig_xlnx.py assist as well need to do changes there as well

lopper/assists/baremetal_bspconfig_xlnx.py
```diff
@@ -37,7 +37,7 @@ def xlnx_generate_bm_bspconfig(tgt_node, sdt, options):
     if options.get('outdir', {}):
         sdt.outdir = options['outdir']
    
-    mem_ranges = get_memranges(tgt_node, sdt, options)
+    mem_ranges, lable_names = get_memranges(tgt_node, sdt, options)
```
onkarharsh, line 40: If the variable is not being used anywhere else in the file and it is defined just to match the return values, please use '_' as the variable name.
e.g. mem_ranges, _ = get_memranges(tgt_node, sdt, options)

This is applicable for all the files

kedareswararao (approved):

lopper/assists/baremetal_bspconfig_xlnx.py
sathishkumar-amd: Added _ as suggested

PR #467, opened 2025-02-11
__init__: Give priority to user defined paths for assists and lops paths search

onkarharsh:
If a user is explicitly passing directories using -I command line option or by setting LOPPER_INPUT_DIRS environment variable, that directory should be given the highest priority over the default lops and assists paths. Otherwise, the purpose of these options get defeated.

### DIFF
--- a/lopper/__init__.py
+++ b/lopper/__init__.py
@@ -388,7 +388,7 @@ class LopperSDT:
         self.support_files = support_files
 
         if self.verbose:
-            search_paths = [ lopper_directory ] + [ lopper_directory + "/assists/" ] + self.load_paths
+            search_paths = self.load_paths + [ lopper_directory ] + [ lopper_directory + "/assists/" ]
             print( "" )
             print( "Lopper summary:")
             print( f"   system device tree: {sdt_files}" )
@@ -684,8 +684,8 @@ class LopperSDT:
             # check the path from which lopper is running, that
             # directory + lops, and paths specified on the command line
             input_file_abs = ""
-            search_paths =  [ lopper_directory ] + [ lopper_directory + "/assists/" ] + \
-                            [ lopper_directory + "/lops/" ] + local_search_paths + self.load_paths
+            search_paths =  self.load_paths + [ lopper_directory ] + [ lopper_directory + "/assists/" ] + \
+                            [ lopper_directory + "/lops/" ] + local_search_paths
             for s in search_paths:
                 # the first found file is the one we return
                 if input_file_abs:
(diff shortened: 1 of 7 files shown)


zeddii:
I have to be really careful here, as this is a change to long released behaviour.

Can you elaborate on the problem you are trying to fix  with the ordering change ?

Lops and assists are not designed to be ordered on a search path level, as that is filesystem dependent. There is a priority property within an assist / lop file that controls the order (versus the search path order).

They (lops and assists) will be queued in their command line order, but after that, it is priority that takes over. If there are lops and assists with the same throughout the system, that is also intended to be handled with priority.

onkarharsh:
Hi @zeddii ,
I am trying to address a situation where a lops/assists file with the same name exists at two different location.
Say there is a lops file (lop-microblaze-riscv.dts) with the same name at two places, one inside the lopper/lopper/lops location which comes with this repository and the other somewhere maintained locally by the user. By default, lopper will check for this lops file within the lopper/lopper/lops which is fair. But, when user is explicitly trying to pick lops from his local path, we should allow them to use it with -I option (or LOPPER_INPUT_DIRS). At present even though -I option is passed by the user indicating to pick lops/assists from his local path, the file is always getting picked up from the default lopper/lopper/lops path.

To add some more background, I am trying to work upon a use case where I can easily plug and play my lops and assists with lopper frontend. The lopper in a binary format with all the lops and assists from this repo is now part of our proprietary tools. Now, as a lopper backend developer (who only cares about lops and assists), I would like to test my lops/assists changes by using -I (or LOPPER_INPUT_DIRS) and send the patches. I don't have to create the binary again for my local testing if only assists or lops are being changed.

zeddii:
Thanks for the description. I assumed it was something like this. 

The built-in priority mechanism won't help with the files of the same name, since as you've noted, lopper will pick the first match and use it.

My original thoughts on files with the same name was to aggregate / overlay them, that allows the inheritance and extension of an existing lop.

That would partially solve what you are trying to achieve, but would require you to add a "noexec;" to the lops to prevent them from running. It also wouldn't solve assists, since they don't function the same way.

Out of curiosity, wouldn't it be preferable (I say preferable, since it is definitely possible) to use a different name for the lops under test, and modify the lopper command line to be clearly using them, versus needing to inspect the search directories (and hence save the command line anyway) to know the ordering ? Or are you always running with the same search paths, and then only creating the override version of the lops when you want to test them, so the command line is identical between the two runs ?

I'm just trying to understand this a bit more, but I am leaning towards merging this (with a tweaked commit message), since the use case is valid, and I don't have another existing mechanism that would fully address this.

onkarharsh:
Hi @zeddii ,
In our workflows, for creating board support packages using SDTs, a sequence of lopper commands are being called and more often we have a wrapper script calling these commands. e.g. gen-machine-conf or the wrapper python scripts for baremetal BSPs. These wrappers then become the interface for the embedded developers and start acting like a tool. It is cumbersome for an embedded developer to go inside these tools and then update the lopper command of their interest and test. Each developer usually own one or two lops/assists out of many available here and forcing them to update the tool with their limited ownership may not be ideal IMHO.

zeddii:
I reworded the commit a bit, but ended up merging it.

While there are other ways to do this, this is simple enough and there's no impact to existing behaviour.

PR #469, opened 2025-02-13
lopper: assists: domain_access: Traverse all domains to retrieve the …

kedareswararao:
…absolute path of each domain

Domains can exist under the default system or at the top level, meaning the subsystem itself can be a domain. Update the logic to handle all possible cases by traversing all nodes and matching the node label or node name with the provided domain name.

### DIFF
--- a/lopper/assists/domain_access.py
+++ b/lopper/assists/domain_access.py
@@ -145,7 +145,8 @@ def core_domain_access( tgt_node, sdt, options ):
     if tgt_node.abs_path == "/":
         if sdt.target_domain:
             try:
-                tgt_node = sdt.tree[sdt.target_domain]
+                target_domain_path = [node.abs_path for node in sdt.tree['/'].subnodes() if node.label == sdt.target_domain or node.name == sdt.target_domain]
+                tgt_node = sdt.tree[target_domain_path[0]]
             except Exception as e:
                 print( f"[ERROR]: target domain {sdt.target_domain} cannot be found" )
                 sys.exit(1)


zeddii:
I've pushed some changes that will allow a domain such as you are looking for to be processed by the assist, even if it isn't a full domain.

But we can't enable searching like you have, since that is only intended for domains, not subsystems.

I've fixed/enabled the ability to pass an absolute path to a target domain, as follows:

 lopper.py -f   --enhanced -t "/domains/Secure" -a domain_access --auto -v  -x '*.yaml' -i domains.yaml system-top.dts test.dts

Even if the target "domain" is a subsystem, the processing will proceed.

The reason that something like -t "APU" works in a domains.yaml when it is a domain, is the yaml transformations. They take care of flipping the name of a yaml domain to the label of a node, and assigning a unique domain@<count> name to the node.  This is only available to nodes that are compatible with "openamp,domain-v1" (note: we could change that to not include "openamp", but it still needs to be a domain).

When the domain_access assist looks up the target domain (what is passed via -t), it used the lopper tree notation of [], that notation automatically searches by node name AND by label. Which is why something like -t "APU" can find a domain called APU.

The searching isn't ideal in any scenario, since it can be ambiguous and find the wrong node. A full path is the safest way to specify a target. This is why it shouldn't be extended like the proposed patch.

Alternately if a full path isn't passed, the yaml expansion lops could provide special processing for subsystems (and do a name, label flip in very specific situations), or another lop could be passed along side the default yaml expansion that allowed the switch of the labels or the conversion to domains. A final option would be to ensure that whatever is generating the domains.yaml generates target "domains" as domains and not subsystems.

PR #493, opened 2025-03-05
lopper: Ensure Only Existing DT Node Parameters

preethi-6:
Added a continue statement in the for loop to skip parameters that do not exist in the DT node. This
prevents unnecessary population of non-existent
parameters in xparameters.h

### DIFF
--- a/lopper/assists/baremetal_xparameters_xlnx.py
+++ b/lopper/assists/baremetal_xparameters_xlnx.py
@@ -1,6 +1,6 @@
 #/*
 # * Copyright (c) 2020 Xilinx Inc. All rights reserved.
-# * Copyright (c) 2024 Advanced Micro Devices, Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


lopper/assists/baremetal_xparameters_xlnx.py
```diff
@@ -334,6 +335,7 @@ def xlnx_generate_xparams(tgt_node, sdt, options):
                         except KeyError:
                             _warning(f"Get property value is failed for {prop} and node is {node.name}, adding default value as {0}")
```
kedareswararao, line 336: this warning message we need to update we should say proprery is not present in the device-tree so not generating define in xparamers.h file same is true for above intr_id case as well

@zeddii : Please don't merge the loppers/assists/b* and lopper/assists/gen_domain_dts changes until me or onkar approves the changes.

PR #500, opened 2025-03-11
baremetalconfig_xlnx : Add support for gmiitorgmii-addr property

venodela:
Add logic to extract the reg property from subnodes where compatible is "xlnx,gmii-to-rgmii-1.0" and set it as gmiitorgmii-addr. Default to 0 if no match is found.


zeddii:
Are any additional approvals required for this ? If not, I'll merge it tomorrow.

PR #501, opened 2025-03-12
gen_domain_dts: update linux ignore list to remove mmi_10gbe and gty

sivadur:
Update linux ignore list to remove mmi_10gne and gty so that these nodes will reflect in corresponding linux device tree.


zeddii:
Without a ack or nack by reviewers, I'll assume this is ok and will merge it tomorrow.

onkarharsh:
Looks Good to Me.

PR #521, opened 2025-03-27
ops/assists: Introduce new assist file to address PL overlay issues

NavaKishoreManne:
To ensure backward compatibility, a new assist file will be created instead of modifying the existing xlnx_overlay_dt implementation. This new file will resolve all identified issues related to PL overlays, including incorrect phandle handling, parsing errors, and other limitations. It will ensure proper overlay handling, maintain references as defined in the SDT-generated pl.dtsi, and improve error messages for better debugging.

Usage: lopper -O <output_dir> -f --enhanced <system-top.dts> <lopper-gen DT>
 -- xlnx_overlay_dt <machine> <config> <PL DTSI (SDT-gen)>

Arguments:
 <output_dir>     - Directory where the lopper generated overlay `pl.dtsi`
                     file will be stored
 <system-top.dts> - Full system device tree source
 <lopper-gen DT>  - Lopper-generated system device tree file(excluding PL node)
 <machine>        - Target architecture: cortex-a9 | cortex-a53
                                         |cortex-a72 | cortex-a78
 <config>         - Configuration type: full | segmented | dfx
                                       | external-fpga-config
 <PL DTSI>        - pl.dtsi file, generated by the SDT tool


zeddii:
I deleted my previous comment, since you already know of the overlay_of() routines and the ability to write overlay files. 

Is there still something wrong with them ? What this routine is doing is far to "hands on" with respect to formatting, etc, and really needs to be replaced in the future.

But you definitely should not have to update the symbols node if --symbols is passed to lopper, is that not working ?

NavaKishoreManne:
HI @zeddii,
Initially, we explored using overlay_of() for generating overlays but encountered challenges in managing phandles. To address this, we referred to the approach outlined in commit: https://github.com/devicetree-org/lopper/commit/d2496b3d5d9ba5e042e9305d0df8083faaaf22b0#diff-728f09c7ced18e76e326c5a94b87387042579721630418b48bdbebe95b1c40e7R46-R48
According to this approach, all phandles must be identified and added to a dictionary. While this method works well for PS IPs, as their configurations remain fixed, handling PL components is more complex. Since PL designs can vary significantly, maintaining a dictionary becomes challenging. For example, in some cases, remote-end-point may contain only a single phandle, whereas in others, it may include multiple references. Unlike fixed properties such as clocks or interrupts, the structure is not uniform, and there is no guaranteed ordering of PL nodes.
To streamline this process, the new assist file ensures that the original node (as generated by the SDT tool) and its phandle references are preserved, eliminating the need to manually maintain a dictionary. This approach enhances consistency and reduces complexity, making overlay generation more efficient and reliable.
We have validated this approach across different designs, including Zynq, ZynqMP, Versal, and Telluride. In all cases, the overlays were generated as expected without any issues.
As you are aware, some critical use cases were failing due to the limitations of the existing implementation. To unblock these scenarios, I have created this new assist file. If a more generic solution becomes available to handle node order and phandle management without relying on a dictionary, we can consider adopting those changes in the future.

zeddii:
I disagree with the complexity comment about the phandle dictionary, if you definitions of the properties are documented, it is a one time update. That dictonary simply tells lopper how to do symbolic replacement when writing source formats. It has nothing to do with phandle consistency

Unfortunately the overlay_of() code will stay as it currently is unlesss it is actively used and new features suggested. 

What you have in that assist is very complex (and unmaintainable imho) as it is actively generating it's own DTS and not using core lopper routines.

Surely you are not describing properties that don't have a standard/consist definition (i.e. remote-end-point), if so, that is unreadable by an operating system can cannot be the case. Lopper can handle variable sized properties with phandles, we just need the description/definition of the property.

I lean towards carrying this assist outside the lopper tree, as it is inconsistent with the overall lopper design and we wouldn't want to see it copied.

I also don't understand your node order comments.  Copied nodes into any overlay are the same as they are in the main transform tree now. Are you seeing something different ? If the node ordering change I made isn't working, let me know, as I'll revert it and return to the previous ordering .. that change was only made to support your request and the overlays.

zeddii:
I have a good idea about how to merge this now, and have a plan to fix it in the futre.

Can you provide me the set of input files and outputs that you tested this assist with ?

If so, I can work to convert the assist to using overlay_of() and the other core lopper routines versus the current line by line parsing and writing of the DTS. And I can ensure that both outputs are identical (or at least syntaxtically and semantically the same).

NavaKishoreManne:
[Uploading for_Bruce.zip…]()
I agree—using the Lopper command overlay_of() is the ideal approach for overlay generation.
 
However, ensuring that all phandles are correctly identified and processed is crucial. Since Lopper does not inherently understand phandles, it relies on a formatting table to handle them.
 
This method works well for PS IPs, where configurations remain fixed. However, PL components introduce complexity, as their designs can vary significantly. Maintaining a dictionary of phandles becomes challenging, requiring continuous updates whenever new references are introduced.
 
The remote-end-point property is just one example of these edge cases. But since it is PL and many combinations of IPs and connections are possible, many unknown phandle references may emerge, and missing entries in the dictionary could lead to inconsistencies and errors. 
 
To address this, I have introduced this new approach with xlnx_overlay_pl_dt.py, which eliminates the need for manually maintaining a dictionary by relying on SDT-generated pl.dtsi files for overlay generation. But i also agree with you on the fact that we shall be using lopper frontend calls for modifying this.
 
For the same reason, I have also created another xlnx_overlay_dt.py to use the Lopper command overlay_of(). While this updated assist works correctly for existing cases, it has not been fully validated across all scenarios. The primary concern remains the dictionary-based phandle resolution, as missing entries may still cause issues. I have attached this updated xlnx_overlay_dt.py assist, along with the expected input and output files for reference. Please review and provide your feedback.

zeddii:
I can't access the zip file mentioned in the comments (or maybe I'm misunderstanding where I should be finding it).

But to unblock progress, I've merged these.

But if someone can send me the mentioned assist that uses overlay_of() and the zip file with the inputs from this assist, I'll raise an issue here, and make sure they produce the same result.

PR #522, opened 2025-03-28
lopper: assists: yaml_bindings: Add tile property

laeachur:
Add new xlnx,tile-formats property for framebuffer read and write.


laeachur:
this pull request is cancelled due to the issue in yaml file.

PR #554, opened 2025-05-27
assist: Exclude BRAM node if ECC is disabled

lakshmaiahm:
Exclude the BRAM node from peripheral test application generation if the 'xlnx,ecc-status-registers' property exists and its value is 0. This avoids generating test examples for ECC-disabled BRAM instances.


onkarharsh:
Please withdraw this Pull Request. We already have property value support in driver YAML bindings.

PR #556, opened 2025-06-03
assists: baremetallinker: Add DDR_PL NOC specific address ranges

lakshmaiahm:
This fix aligns the address with the expected value used in Classic platform definitions and avoids negative address mapping issues.


onkarharsh:
Looks good to me

PR #562, opened 2025-06-18
lopper:assists: Add support to read the custom configurations in the …

sathishkumar-amd:
…YAML

Add support for reading depends from custom configurations in YAML files. If a YAML file contains a condition within the depends section in the form of a Python code block, read and execute this code to update the depends based on the specified conditions. This approach allows dynamic configuration adjustments based on predefined logic, improving flexibility and customization in managing depends


onkarharsh (approved):

onkarharsh:
Looks good to me.

lopper/assists/bmcmake_metadata_xlnx.py
```diff
@@ -218,7 +218,22 @@ def generate_hwtocmake_medata(sdt, node_list, src_path, repo_path_data, options,
...
+    if meta_dict.get("condition") and family:
+        match_cpunode = bm_config.get_cpu_node(sdt, options)
+        proc_ip_name = match_cpunode.propval('xlnx,ip-name')[0]
+        local_scope={
+            "proc":proc_ip_name,
+            "platform":family,
+            "depends":[]
+        }
+        try:
+            exec(meta_dict["condition"], {}, local_scope)
```
zeddii, line 232: What are the possible types of commands that could be run ? If it is just a condition, is there any checking to make sure that it is only a condition ?

When I was implementing the lop capability to run python code, I jumped through several steps to take away potentially harmful libraries (system, os, etc).

Just trying to make sure sure that no damage could be caused to the machine running this, if something incorrect gets into the yaml condition.

sathishkumar-amd: The condition should be a Python Boolean expression or minimal inline logic that checks values such as processor type, platform, etc.
For example:
`condition: |
  depends = ["abc"]
  if platform == "zynq":
      depends.append("abc1")
  elif platform == "versal":
      depends.append("abc12")`

However, there is currently no explicit restriction, so it could technically contain dangerous code (e.g., import os; os.system("rm -rf /")) if misused.

zeddii: Right. I agree that there's no issue with the intent of how it is to be used. There is just the possibility it could be misused.

Have you considered some of the common patterns for limiting the scope of what is available before executing the parsed code ?

There's one example in tree.py in lopper (exec_cmd) that attempts to make the code safe and limit what can be done. It is not perfect by any means, but if we put some limitations in place, at least we know that we've tried and can improve it later if something does come up as a vulnerability.

PR #564, opened 2025-06-20
lopper: Fix tx_clk index handling in clock property

shubhraamd:
The clocks property in the device tree consists of tuples in the format: <phandle clock-id>. To correctly extract the clock ID associated with "tx_clk",
the offset needs to account for the two-cell structure.

### DIFF
--- a/lopper/assists/baremetalconfig_xlnx.py
+++ b/lopper/assists/baremetalconfig_xlnx.py
@@ -242,7 +242,8 @@ def get_clock_offset(node):
     if node.propval("clock-names") != ['']:
         clock_names = node["clock-names"].value
         try:
-            tclk_offset = clock_names.index("tx_clk") + 1
+            tclk_offset = clock_names.index("tx_clk")
+            tclk_offset = 2 * tclk_offset + 1
         except ValueError:
                 tclk_offset = 1
 


lopper/assists/baremetalconfig_xlnx.py
```diff
@@ -242,7 +242,8 @@ def get_clock_offset(node):
     if node.propval("clock-names") != ['']:
         clock_names = node["clock-names"].value
         try:
-            tclk_offset = clock_names.index("tx_clk") + 1
+            tclk_offset = clock_names.index("tx_clk")
+            tclk_offset = 2 * tclk_offset + 1
```
kedareswararao, line 246: what exactly you are trying to get here? index of the tx_clk value in the clock-names property?

shubhraamd: Hi Kedar,

Yes, the intent here is to get the index of "tx_clk" in the clock-names property, but since the clocks property consists of tuples in the form of <phandle clock-id> (i.e., 2 cells per clock), we need to compute the correct offset into the flattened clocks array.

So:

clock_names.index("tx_clk") gives us the position in clock-names.

Since each clock takes 2 cells, the clock-id is at 2 * index + 1.

kedareswararao: got it thanks for the clarification

kedareswararao (approved):

PR #566, opened 2025-06-23
lopper/xlnx_overlay_pl_dt: Add support to override firmware-name via …

NavaKishoreManne:
…command-line

This patch introduces a new optional command-line argument `--firmware-name=<name>` to the xlnx_overlay_pl_dt assist. When provided, this overrides the default `firmware-name` property in the generated overlay DTS file (`pl.dtsi`).

This enhancement is useful in scenarios where a custom bitstream name needs to be embedded into the overlay generation flow, especially when the firmware is managed externally.

Usage example:
  lopper ... -- xlnx_overlay_pl_dt <machine> <config> <pl.dtsi>
--firmware-name=my_custom.bit


zeddii:
Looks fine to me, any extra reviewers required for this one ?

kedareswararao (approved):

PR #575, opened 2025-07-19
assists: baremetallinker_xlnx.py: Add missing memory region for NoC_C…

lakshmaiahm:
…2_C3

The base address 0xA00000000 used by the NoC_C2_C3 memory region was not present in the valid memory regions list within the baremetallinker assist. Due to this, the linker was incorrectly assigning the region to an adjacent block (e.g., DDR_LOW_1), resulting in incorrect memory address and size calculations.

This patch adds the missing entry "0xA00000000": "DDR_LOW_1A" to the memory map, ensuring the correct association of the NoC_C2_C3 region during linker script generation.


zeddii:
Any other reviewers required ? this looks fine to me.

lakshmaiahm:
This region (0x8_0000_0000 - 0x5_0000_0000) is a single DDR Memory IP 
Hence these changes are not needed and closing this PR

PR #581, opened 2025-07-30
assists: bmcmake_metadata: Add support to define generic CMAKE_MACHINE defines using family

onkarharsh:
For every new device, a new CMAKE_MACHINE/SUBMACHINE and the corresponding macro names are required to distinguish the cmake and the C code in baremetal. Add support to define these variables based on the family entry in SDT. This will avoid manual intervention in the BSP code for each new device.

### DIFF
--- a/lopper/assists/bmcmake_metadata_xlnx.py
+++ b/lopper/assists/bmcmake_metadata_xlnx.py
@@ -324,14 +324,41 @@ def generate_hwtocmake_medata(sdt, node_list, src_path, repo_path_data, options,
             if sdt.tree['/'].propval('board') != ['']:
                 val = sdt.tree['/'].propval('board', list)[0]
                 fd.write(f'set(BOARD "{val}" CACHE STRING "BOARD")\n')
+
+            # Helpful in distinguishing minor differences within the same family device
+            # e.g. spartanuplus-spartanuplusaes, Versal_2ve_2vm (full and small variants)
+            variant_list = sdt.tree['/'].propval('variant', list)
+            if variant_list[0]:
+                fd.write(f'set(VARIANT {";".join(variant_list)})\n')
+
+            cmake_machine = ""
+            cmake_submachine = ""
+
             match_cpunode = bm_config.get_cpu_node(sdt, options)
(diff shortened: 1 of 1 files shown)


lopper/assists/bmcmake_metadata_xlnx.py
```diff
@@ -324,14 +324,41 @@ def generate_hwtocmake_medata(sdt, node_list, src_path, repo_path_data, options,
             if sdt.tree['/'].propval('board') != ['']:
                 val = sdt.tree['/'].propval('board', list)[0]
                 fd.write(f'set(BOARD "{val}" CACHE STRING "BOARD")\n')
+
+            # Helpful in distinguishing minor differences within the same family device
+            # e.g. spartanuplus-spartanuplusaes, Versal_2ve_2vm (full and small variants)
+            variant_list = sdt.tree['/'].propval('variant', list)
```
kedareswararao, line 330: This variant property always present? we need to check for existence of this property, right?

lopper/assists/bmcmake_metadata_xlnx.py
```diff
@@ -324,14 +324,41 @@ def generate_hwtocmake_medata(sdt, node_list, src_path, repo_path_data, options,
...
             match_cpunode = bm_config.get_cpu_node(sdt, options)
-            if re.search("microblaze", match_cpunode['compatible'].value[0]):
-                if match_cpunode.propval('xlnx,family') != ['']:
-                    family = match_cpunode.propval('xlnx,family', list)[0]
-                    # Special handling for Versal platform update the variable to inline with PS processors.
-                    if family == "versal":
-                        family = "Versal"
-                    fd.write(f'set(CMAKE_MACHINE "{family}" CACHE STRING "CMAKE MACHINE")\n')
+
+            family = sdt.tree['/'].propval('family', list)
```
kedareswararao, line 339: same comment here as well...

kedareswararao (approved):

onkarharsh:
I have considered below scenarios:

PS + softMB : CMAKE_MACHINE is expected to be the PS family name
Pure MB: CMAKE_MACHINE should be MB family name
PS : Family name

PR #588, opened 2025-07-31
Lopper: Assists: Allow application creation when required memory is e…

preethi-6:
…xactly matched

Updated the memory validation logic in check_for_mem to allow application creation when the available memory size is equal to the required size. Earlier, the validation passed only if available memory was greater than the required size. This caused memory_tests to fail even when exactly 0x2000 bytes were allocated, as defined in the YAML. Changed the condition from greater than to greater than or equal to, so that exact matches are also accepted.

### DIFF
--- a/lopper/assists/baremetal_validate_comp_xlnx.py
+++ b/lopper/assists/baremetal_validate_comp_xlnx.py
@@ -56,7 +56,7 @@ def check_for_mem(sdt, options, mem_type, required_mem):
             if required_mem_start != 0x0:
                 if (required_mem_start == start) and (size <= required_mem_size):
                     return False, mem_type, required_mem_size
-            if size > required_mem_size:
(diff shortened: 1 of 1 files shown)


lopper/assists/baremetal_validate_comp_xlnx.py
```diff
@@ -56,7 +56,7 @@ def check_for_mem(sdt, options, mem_type, required_mem):
             if required_mem_start != 0x0:
                 if (required_mem_start == start) and (size <= required_mem_size):
```
onkarharsh, line 57: Update this entry as well

zeddii:
Any updates on this ? Is it still valid ?

onkarharsh (approved):

onkarharsh:
Change looks good now. Please proceed to merge this. Thanks.

PR #596, opened 2025-08-11
openamp: Linux: Remove non-openamp IPI nodes if OpenAMP domains are present.

bentheredonethat:
In Linux there are now all APU-associated IPIs present.

In case of Remoteproc being present remove all IPI MBOX nodes nodes as we cannot tell if there will be conflict.

If there is RPMsg present, only remove IPIs that conflict for the IPC purpose.


bentheredonethat:
Hi @zeddii  please wait  on merge.

I need review on this first.

PR #597, opened 2025-08-12
assist:baremetal_getsupported_comp_xlnx: Add variant support for plat…

sathishkumar-amd:
…form-specific example filtering

Enhance baremetal component discovery to support variant-based platform filtering. This update adds a variant parameter to the get_yaml_data() function and extracts variant information from the SDT root node alongside the existing family data. The platform matching logic has been improved to check both family and variant values using a combined family+variant list for more accurate component filtering.


onkarharsh:
Looks good to me

PR #602, opened 2025-08-16
assists/subsystem: fix cpu expand check

bentheredonethat:


bentheredonethat:
@zeddii  ill be pushing at least one more commit fix on this

i dont know what the 'proper' fixes are for these but these unblock immediate issues

zeddii:
access-json is very much needed. It can't be removed.

Send me your current commands and I'll reproduce it locally.

yaml/json source properties should be properly escaped, so this is slipping through somehow

bentheredonethat:
@zeddii  i have removed that commit in this brnch

will send you command, yaml and SDT

bentheredonethat:
@zeddii  i have added some spot fixes in spots that should not affect the given lopper run.

I dont know where the 'most clean' places are to fix each so I added them as I saw them. thanks.

PR #604, opened 2025-08-19
baremetallinker_xlnx: Select correct DDR system map for platform

sivadur:
Newer variants of versal platform can also have axi noc2 IPs and it still complies with the DDR address map of Versal, so assign and use the correct ddr map for Versal platforms with axi noc2.


kedareswararao (approved):
Changes looks fine to me

PR #611, opened 2025-08-21
lopper:assist: Define FRL enable macro based on HDMI mode

gugulot:
Define a macro to enable or exclude FRL-related logic in the driver, based on the HDMI mode configured in the HDMI 2.1 Rx subsystem IP.


zeddii:
Any other reviews needed ?

onkarharsh (approved):

PR #613, opened 2025-08-21
docs: Add docs directory for documenting AMD specific lops and assists usages

onkarharsh:
Add docs directory for documenting AMD specific lops and assists usages. Add linux related usages to begin the documentation process.


zeddii:
I'm good with the structure, but we should, in a similar way to the specification/ documentation, have a Makefile that builds the docs into various formats via sphinx.

That will keep all the documentation consistent, I plan on converting the main READMEs in a similar manner.

onkarharsh:
Thanks @zeddii. I have converted the README to rst file format satisfying the sphinx requirements.

PR #628, opened 2025-09-09
gen_domain_dts: Prune IPI node properties in Linux DT

Harini-T:
Add xlnx,zynqmp-ipi-mailbox.yaml to the yaml bindings to prune the additional properties of IPI nodes that are not required for Linux DT.

Add logic to prune IPI child node properties in the Linux DT according to the Yaml Schema by retaining only required properties specified in the Yaml.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -382,11 +382,14 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
         driver_compatlist = []
         # Shouldn't delete properties
         driver_proplist = ["#interrupt-cells", "#address-cells", "#size-cells", "device_type"]
+        ipi_schema = None
         for yaml_prune in yaml_prune_list:
             yaml_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), yaml_prune)
             schema = utils.load_yaml(yaml_file)
             driver_compatlist = driver_compatlist + compat_list(schema)
             driver_proplist = driver_proplist + schema.get('required',[])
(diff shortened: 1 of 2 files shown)


Harini-T:
@kedareswararao @bentheredonethat 
Can you please review the PR?

bentheredonethat:
review comment - why is there special handling for this? shouldn't schema check and handling do this? we dont do this for versal mailbox nodes right?

Harini-T:
This handling is specific to Linux DT as the DTB size increased because of the presence of the IPI nodes causing PDI boot failure.
The existing handle prunes only the parent node properties.
Hence added the logic to prune child node properties for IPI as well.

bentheredonethat:
LGTM @zeddii  though i wonder if there is a way to make this a list of YAML schema and based on them do prune - similar to how zephyr does it

zeddii:
So I understand, you mean a way to have either nodes or properties in a yaml input and then rather than merging and adding them, they would be removed ?

That's definitely something that could be done. 

I just need to think of a notation that wouldn't be complex.

In the lop files, I did it with "!" to negate the operation.

So something like:

mycpucluster:
    - !property-to-delete
    
But using a syntax that is valid yaml.

Is that what you mean ?

kedareswararao:
> LGTM @zeddii though i wonder if there is a way to make this a list of YAML schema and based on them do prune - similar to how zephyr does it

@bentheredonethat: In the case of Linux or Zephyr, we should be reading from the corresponding bindings folder instead of hardcoding YAMLs in the Lopper repo. This is already in plan for the next release.

PR #629, opened 2025-09-09
lopper: assists: gen_domain_dts: save nodes in domain's access list

bentheredonethat:
Currently the nodes saved in a non-Linux gen-domain run comprise only of nodes mentioned in the address map.

If the input SDT also has a domain that maps to the target CPU and this domain has an access list, then also save those nodes from being deleted.

As now a user can describe TCMs in a domain access list this also means that a previous hard-coding to save TCMS is no longer needed as the domain YAML can be used to save it.


bentheredonethat:
@kedareswararao  @onkarharsh  @zeddii  please review

PR #643, opened 2025-10-08
lopper:assist:zephyr: Add support to generate AXI-Ethernet Node

Tarun-Maddipatla:
Add support to pull in dma-channels property for AXI-DMA Node


Tarun-Maddipatla:
There was a mistake in the subject of the above patch. Will create a new patch and pull request again

PR #733, opened 2026-04-09
gen_domain_dts: Extend mapped_nodelist to include bus children nodes

harishpamarthi:
Preserve I2C/SPI device nodes when their parent bus is mapped to CPU cluster After building mapped_nodelist, iterate all mapped nodes and add their children to the preservation list. This ensures devices like RTC on i2c1 bus are kept in the final device tree output


harishpamarthi:
@onkarharsh Please help with review.

PR #734, opened 2026-04-10
assist: gen_domain_dts: Add support for PS processors accessing PL peripherals

dbingi-amd:
PS processors can access PL peripherals mapped in their address space, but the existing Lopper logic only generated PL nodes for pure-PL designs. This change extends the Zephyr domain DT generation to retain and correctly transform PL peripheral nodes under amba_pl for PS+PL designs as well.

- Preserve PL nodes under amba_pl in xlnx_remove_unsupported_nodes() when they match a Zephyr-supported compatible string, applying the same property pruning and clock generation as PS nodes.

- Preserve interrupt-parent and interrupts properties for PL nodes whose parent bus lacks a bus-level interrupt-parent, ensuring correct interrupt routing (direct-to-GIC or via AXI INTC).

- Preserve clock-frequency and xlnx,count-width for AXI Timer nodes instead of requiring a clocks phandle, matching how the Zephyr counter driver consumes these properties.

- Extract shared PL peripheral transformations (AXI DMA dma-channels, AXI Ethernet subsystem restructuring, AXI SPI address/size cells, UARTNS550 current-speed) into _apply_pl_peripheral_transforms() helper, eliminating code duplication between the PS+PL path (xlnx_remove_unsupported_nodes) and pure-PL path (xlnx_generate_mb_board_dts).

- Remove empty amba_pl nodes when no PL peripherals with reg properties are present, preventing unnecessary bus nodes in PS-only CED designs.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -750,13 +750,35 @@ def xlnx_generate_zephyr_domain_dts_arm(tgt_node, sdt, options, machine):
                                 for cell in intr_list[i:i+3] + [0xa0]]
             elif node.propval('interrupts') != ['']:
                 intr_list = node["interrupts"].value
-                intr_list.append("0xa0")            
+                intr_parent_cells = None
+                try:
+                    ip_val = node.propval('interrupt-parent')
+                    if ip_val != ['']:
+                        ip_node = sdt.tree.pnode(ip_val[0])
+                        if ip_node is None:
+                            # Dangling phandle (e.g. imux deleted) - repoint to GIC
+                            for gic_candidate in sdt.tree['/'].subnodes():
+                                if gic_candidate.propval('compatible') != ['']:
+                                    if any('arm,gic' in c for c in gic_candidate.propval('compatible', list)):
+                                        node['interrupt-parent'].value = [gic_candidate.phandle]
+                                        break
+                        elif ip_node.propval('#interrupt-cells') != ['']:
+                            intr_parent_cells = ip_node.propval('#interrupt-cells', list)[0]
+                except Exception:
+                    pass
+                if intr_parent_cells is not None and intr_parent_cells <= 2:
+                    pass
+                elif len(intr_list) >= 3 and len(intr_list) % 3 == 0:
+                    node["interrupts"].value = [cell for i in range(0, len(intr_list), 3)
+                                    for cell in intr_list[i:i+3] + [0xa0]]
 
             if compatible == "cpus,cluster":
                 node.name = "cpus" 
 
-            # Collect all the wwdt nodes
+            # Collect all the wwdt nodes (PS WWDT and PL timebase WDT)
             if any(version in node["compatible"].value for version in ("xlnx,versal-wwdt-1.0", "xlnx,versal-wwdt")):
+                wwdt_nodes.append(node)
+            if any(version in node["compatible"].value for version in ("xlnx,axi-timebase-wdt-3.0", "xlnx,xps-timebase-wdt-1.00.a")):
                 wwdt_nodes.append(node)
             if "amd,versal2-ufs" in node["compatible"].value:
                 ufs_nodes.append(node)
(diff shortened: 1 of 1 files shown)


dbingi-amd:
Hi @kedareswararao,

please review.

lopper/assists/gen_domain_dts.py
```diff
@@ -1088,17 +1184,42 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine):
...
+                        if is_timer:
+                            if 'clocks' in required_prop:
+                                required_prop.remove('clocks')
+                            for p in ('clock-frequency', 'xlnx,count-width'):
+                                if p not in required_prop and node.propval(p) != ['']:
+                                    required_prop = required_prop + [p]
                         # Create fixed clock nodes
                         if 'clocks' in required_prop:
-                            if any(clock_prop := (re.search(r'xlnx,.*-clk-freq-hz$', prop)) for prop in prop_list):
+                            if any(clock_prop := (re.search(r'xlnx,.*-clk-freq-hz(-d)?$', prop)) for prop in prop_list):
```
kedareswararao: why this change required? it's unrelated to commit

lopper/assists/gen_domain_dts.py
```diff
@@ -1088,17 +1184,42 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine):
...
+                                clk_freq = 0
+                                try:
+                                    first_clk_ph = node.propval('clocks')[0]
+                                    clk_node = sdt.tree.pnode(first_clk_ph)
+                                    if clk_node and clk_node.propval('clock-frequency') != ['']:
+                                        clk_freq = clk_node.propval('clock-frequency')[0]
+                                except Exception:
+                                    pass
                             else:
-                                # If there is no clk-freq property use 0MHZ as default this prevent
```
kedareswararao, line 1100: why to remove these comments?

lopper/assists/gen_domain_dts.py
```diff
@@ -1662,14 +1762,6 @@ def xlnx_generate_zephyr_domain_dts(tgt_node, sdt, options):
                             if node.propval('reg-shift') != ['2']:
                                node["reg-shift"] = LopperProp("reg-shift")
                                node["reg-shift"].value = 2
-                            if node.propval('current-speed') == ['']:
```
kedareswararao, line 1665: Reason to remove this code?

lopper/assists/gen_domain_dts.py
dbingi-amd: this (-d)? regex change is a minor robustness improvement, not a strict requirement for the PS+PL feature. The improved clock phandle resolution (Level 3 fallback added in this commit) already handles the case. I'll drop this regex change from the commit if not required.

lopper/assists/gen_domain_dts.py
dbingi-amd: The comment was not intentionally removed. I would like to add the comment if preferred.

lopper/assists/gen_domain_dts.py
dbingi-amd: This code is not removed -- it was refactored into the shared _apply_pl_peripheral_transforms() helper (line 845-853) to avoid duplication between the PS+PL and pure-PL paths.

lopper/assists/gen_domain_dts.py
kedareswararao: okay please drop

lopper/assists/gen_domain_dts.py
kedareswararao: please add it back

lopper/assists/gen_domain_dts.py
kedareswararao: okay fine

lopper/assists/gen_domain_dts.py
dbingi-amd: File was updated with the changes

lopper/assists/gen_domain_dts.py
dbingi-amd: comments are added back

kedareswararao (approved):

PR #736, opened 2026-04-13
lopper:assists: Add additionalProperties support for MicroBlaze CPU

sushil-dev-amd:
Add support for reading additionalProperties from MicroBlaze CPU YAML schema, similar to how it is handled for driver YAML files. This allows optional properties like xlnx,fault-tolerant to be defined in the YAML additionalProperties section and have their corresponding macros generated in xparameters.h when present in the device tree.


onkarharsh (approved):
Looks good to me

PR #738, opened 2026-04-14
Add support for oneOf/anyOf property in yaml

harishpamarthi:
Enhanced baremetal_validate_comp_xlnx.py and common_utils in Lopper to handle dependency validation with below three modes:
1. Traditional format (allOf) - existing flow, all drivers required
2. oneOf - at least one driver from the list must be present
3. anyOf - at least one driver from the list must be present


zeddii:
Any additional reviews needed for this ?

zeddii:
This looks like it was pushed to another PR branch, which I merged. Close this PR if that is the case.

PR #739, opened 2026-04-16
domain_access: apply /domains/<domain>/nodes properties to resolved SDT nodes

bentheredonethat:
domain_access: treat /domains/<domain>/nodes entries as patch specs by resolving their translated absolute paths in the main tree and copying their properties onto the matching SDT nodes

Note label matching is not supported.


zeddii:
The sigils conditional syntax supersedes this PR. closing it.

PR #746, opened 2026-04-28
lopper: assists: xlnx_overlay_pl_dt: exclude address-map property from overlay

arthokal:
**Current status**:
With latest changes in master branch,
When overlay_of() is called, fragments are now automatically generated for any base tree properties that reference overlay nodes. This ensures cross-tree references remain valid. Hence address-map properties are generated in the overlay which provides the complete address-map including both PS and PL device entries.

**Fix**:
address-map is an SDT-specific property that is not consumed by the Linux kernel at runtime. Automatically exclude it from the generated pl.dtsi and clean up any empty overlay fragments left behind.

PR #748, opened 2026-04-30
lopper: assists: xlnx_overlay_pl_dt: exclude address-map property from overlay

arthokal:
**Current status**:
With latest changes in master branch,
When overlay_of() is called, fragments are now automatically generated for any base tree properties that reference overlay nodes. This ensures cross-tree references remain valid. Hence address-map properties are generated in the overlay which provides the complete address-map including both PS and PL device entries.

**Fix**:
address-map is an SDT-specific property that is not consumed by the Linux kernel at runtime. Automatically exclude it from the generated pl.dtsi and clean up any empty overlay fragments left behind.

Closed [PR #746](https://github.com/devicetree-org/lopper/pull/746) as it was based on old upstream master.

@zeddi,
I confirm that PL overlay use cases pass on master and there are no warnings. Please review.
Should I still keep cleaning up of empty fragments through regex? You were mentioning this last time:

> so the check for empty property is not used, but I guess it is a good safety net if something breaks.

CC: @onkarharsh , @kedareswararao , @sivadur

### DIFF
--- a/lopper/assists/xlnx_overlay_pl_dt.py
+++ b/lopper/assists/xlnx_overlay_pl_dt.py
@@ -734,6 +734,9 @@ def write_output_files(overlay_tree, sdt, pl_file, dtso_file):
     # Replace interrupt-parent references from imux to gic
     content = content.replace('interrupt-parent = <&imux>;', 'interrupt-parent = <&gic>;')
 
+    # Remove empty overlay fragments left after property exclusion
+    content = re.sub(r'\n&\w+ \{\n\};\n', '', content)
+
     with open(pl_file, "w") as f:
         f.write(content)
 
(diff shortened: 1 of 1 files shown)


zeddii:
It's ok to keep the check, it just won't be used .. but if something breaks it is a good secondary check.

I have no issues with the change now.

zeddii:
Any other reviews needed for this ?

kedareswararao (approved):

zeddii:
I had to update master to fix some issues with the previous release, can you rebase and resend this pull request ?

arthokal:
@zeddii ,

> I had to update master to fix some issues with the previous release, can you rebase and resend this pull request ?

Fixed. Please review.

PR #749, opened 2026-04-30
yaml_to_dts_expansion: fix ranges boolean property in reserved-memory

Rajukumar45:
When ranges: true is specified in YAML, the bool_as_int config causes it to be stored as [1], resulting in ranges = <0x1> in DTS output instead of the correct empty property ranges;

Fix by adding ranges to RESERVED_MEMORY_BOOLEAN_PROPS and applying the boolean-to-empty fixup on both the parent reserved-memory node and its children. Refactor duplicate logic into a shared _fixup_boolean_props helper.


zeddii:
This is already fixed in master. see commits: ff48f7cf and 7e2114f2

Wherever possible, property name triggers are avoided, so the solution is more generic in master.

If you've found a case that master is missing, send me the inputs and I'll investigate more.

PR #753, opened 2026-05-04
base: phandle_safe_name: replace all invalid label characters

Rajukumar45:
DTS labels must be valid C identifiers ([a-zA-Z_][a-zA-Z0-9_]*), but node names can contain characters from the set [0-9a-zA-Z,._+-].

Previously, phandle_safe_name() only replaced '@', '-', and '/'. This left ',' , '.' and '+' in the generated label, producing invalid DTS output (e.g., "pcie0,0:" for node pcie@0,0).

Add replacements for ',', '.', and '+' so that all non-identifier characters in node names are converted to '_'.


zeddii:
sorry to make work, but my screw up on master has made this conflict. Can you rebase it onto current master and resend. There's a non FF merge causing issues (me fixing things)

zeddii:
This wasn't rebased or resubmitted from what I can see.

PR #755, opened 2026-05-06
baremetal: Add lpddrmc to valid memory IP list

drompich:
Add lpddrmc to list of valid memory IPs for baremetal


sathishkumar-amd (approved):
Looks good to me

PR #756, opened 2026-05-06
baremetalconfig_xlnx: fix clocks single-cell offset; optional prop first int

harishpamarthi:
baremetalconfig_xlnx: fix clock cell indexing and optional CMake ints
  Clocks:
  - With no clock-names, use offset 0 when clocks has a single phandle cell (avoids out-of-range access and spurious warnings on IPs like MDM).
  - On out-of-range offset, return value[0]; warn only when multiple cells exist. Fixes incorrect use of value[1] when only one cell is present. Optional driver metadata (DRIVER_OPTPROP_*):
  - Take node[prop].value[0]; if it is a string, parse the first comma-separated token as int (e.g. "0 , 1 , 0" -> first value only).

### DIFF
--- a/lopper/assists/baremetalconfig_xlnx.py
+++ b/lopper/assists/baremetalconfig_xlnx.py
@@ -237,7 +237,6 @@ def scan_ranges_size(range_value, ns):
     return addr, size
 
 def get_clock_offset(node):
-    tclk_offset = 1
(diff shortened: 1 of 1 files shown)


zeddii:
any additional reviews needed for this one ?

kedareswararao (approved):

PR #759, opened 2026-05-11
gen_domain_dts: Add riscv,timer node generation

drompich:
Add riscv,timer node generation for Microblaze-RISCV


drompich:
@kedareswararao @sivadur Please check and confirm the above changes

kedareswararao (approved):

PR #761, opened 2026-05-15
overlay: fix child_nodes deserialization in multi-pass lopper invocations

arthokal:
**Issue**:
When an overlay file with child nodes (e.g., updated_zocl.dtsi containing zyxclmm_drm under &amba_pl) is passed via -i in pass 1, Lopper embeds the overlay data into the output DTS for preservation across passes. In pass 2 (e.g., gen_domain_dts), Lopper crashes at startup while reconstructing the embedded overlay data:

AttributeError: 'collections.OrderedDict' object has no attribute 'append'

**Root Cause**:
_deserialize_overlay_node() used node.child_nodes.append(child), but child_nodes is an OrderedDict, not a list.

**Fix**:
Changed to node.child_nodes[child.abs_path] = child with explicit parent assignment, consistent with how child_nodes is managed in tree.py.

**Example Lopper command invocations**:
```
LOPPER_DTC_FLAGS="-b 0 -@" ./lopper.py -O lopper_out -f --enhanced -i updated_zocl.dtsi apu_linux-chosen.dts system-top-no-pl.dts -- xlnx_overlay_pl_dt psv_cortexa72_0 full --firmware-name=gmio_async_xrt.pdi

LOPPER_DTC_FLAGS="-b 0 -@" ./lopper.py -O lopper_out/ -f --enhanced -i lop-a72-imux.dts lopper_out/system-top-no-pl.dts apu_linux.dts -- gen_domain_dts psv_cortexa72_0 linux_dt
```

### DIFF
--- a/lopper/__init__.py
+++ b/lopper/__init__.py
@@ -584,7 +584,8 @@ def _deserialize_overlay_node(data, parent=None, tree=None):
 
     for child_data in data.get("children", []):
         child = _deserialize_overlay_node(child_data, parent=node, tree=tree)
-        node.child_nodes.append(child)
+        child.parent = node
+        node.child_nodes[child.abs_path] = child
 
     return node
 


arthokal:
@onkarharsh, please review.

onkarharsh:
Looks good to me

zeddii:
I need a bit more information. While the commit looks ok, the commit message doesn't carry enough information.

What attribute error ? I assume something is accessing .child, or similar, but that needs to be in the commit message.

What are the lopper command lines used to reproduce the issue, as well as the inputs. Both of those are required, since the overlay selftest are currently passing, and I need to check against those as well as to consider alternate implementations (we may not need the serialization anymore)

arthokal:
Hi @zeddii,

Thank you for reviewing. Commit message and [PR description](https://github.com/devicetree-org/lopper/pull/761#issue-4451569544) were updated.

Please review.

zeddii:
Thanks @arthokal — fix is correct. I made one small adjustment before merging: rather than setting \`child.parent = node\` after the recursive call (where the \`parent\` parameter was passed in but never used inside the function body), I moved it to \`node.parent = parent\` at the top of \`_deserialize_overlay_node\`. This establishes the parent backref before any property or child processing, which is more defensive against any code that walks \`.parent\` during construction. You're still listed as author.

Also added a second commit with self-tests covering the crash path and the invariants the fix depends on:

- Fix commit: https://github.com/devicetree-org/lopper/commit/3822a36e72d197c0c2ef60749b2abc059f2346ce
- Tests commit: https://github.com/devicetree-org/lopper/commit/bc2d6dbfbe3f6c7d1f0cba5f1a16ef0a8f8d6a8b

zeddii:
The reason this slipped through existing tests: all our overlay test fixtures use flat overlay nodes (e.g. \`/axi/timer@f1e90000\` with only properties, no child nodes), so the \`for child_data in data.get("children", [])\` loop in \`_deserialize_overlay_node\` was never entered and the \`append()\` call was never reached. The new regression tests use a nested node (\`zyxclmm_drm\` under \`amba_pl\`) to exercise that path directly.

PR #762, opened 2026-05-18
lopper:assist:baremetal_validate_comp_xlnx: Fix srec_spi_bootloader b…

sathishkumar-amd:
…uild failure on Spartan US+ OSPI-enabled design

The baremetal_validate_comp_xlnx assist previously issued a hard validation error when the xilsfl dependency (introduced via srec_spi_bootloader) could not be resolved on MicroBlaze and MicroBlaze RISC-V targets that are not Spartan UltraScale+, even though xilsfl is intentionally unsupported on these platforms, resulting in unnecessary build failures this change introduces an optional fourth argument (parent_app) to enable application-aware validation without impacting existing call sites, enhances platform detection by extracting family and variant from the SDT root node, and updates the validation logic to suppress the xilsfl error and emit a warning only when the parent application is srec_spi_bootloader running on non-Spartan UltraScale+ MicroBlaze or MicroBlaze RISC-V targets, while preserving the existing error behavior for all other platforms and application

### DIFF
--- a/lopper/assists/baremetal_validate_comp_xlnx.py
+++ b/lopper/assists/baremetal_validate_comp_xlnx.py
@@ -99,6 +99,7 @@ def xlnx_baremetal_validate_comp(tgt_node, sdt, options):
     proc_name = options['args'][0]
     src_path = utils.get_abs_path(options['args'][1])
     repo_path_data = options['args'][2]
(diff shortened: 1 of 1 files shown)


sathishkumar-amd:
Hi @onkarharsh,

Please review this

onkarharsh:
Looks good to me

PR #763, opened 2026-05-18
Add per-hart interrupts for RISCV

drompich:
Add per-hart interrupts (riscv,timer and riscv,cpu-intc) for RISCV


drompich:
@onkarharsh  @sivadur Please review this change

onkarharsh:
Looks good to me

PR #764, opened 2026-05-18
assists: xlnx_overlay_pl_dt: Move amba_pl to amba rename after overlay tree construction

arthokal:
**Issue**:
When loading the generated overlay DT on the target, the kernel reports:
`OF: resolver: node label 'amba_pl' not found in live devicetree symbols table`

**Root Cause**:
This happens because lopper renames `amba_pl` to `amba` early in the flow, before the overlay tree is fully built. The generated overlay then contains inconsistent references that the kernel's OF resolver cannot resolve.

**Fix**:
Move the `amba_pl` to `amba` rename to happen after `build_overlay_tree()` completes, so all references to `amba_pl` remain valid during overlay tree construction.

**Example Lopper command invocation**:
`LOPPER_DTC_FLAGS="-b 0 -@" ./lopper.py -O outdir -f --enhanced -i /proj/ssw_xhd/drivers/onkarh/to_aravind/updated_zocl.dtsi /proj/ssw_xhd/drivers/onkarh/to_aravind/apu_linux-chosen.dts system-top-no-pl.dts -- xlnx_overlay_pl_dt psv_cortexa72_0 full --firmware-name=gmio_async_xrt.pdi`

### DIFF
--- a/lopper/assists/xlnx_overlay_pl_dt.py
+++ b/lopper/assists/xlnx_overlay_pl_dt.py
@@ -403,7 +403,7 @@ def prepare_amba_node(amba_node, sdt, tgt_node):
         _warning(f"No aliases node found in device tree")
 
     # Rename the new node in the overlay
-    new_amba_node.name = "&amba"
+    new_amba_node.name = "&amba_pl"
     new_amba_node.label = ""
 
     # Delete structural properties that should only exist in the base device tree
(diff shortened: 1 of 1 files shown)


arthokal:
@onkarharsh , @kedareswararao 
Please review.

onkarharsh:
Looks good to me, it is co-authored.

PR #765, opened 2026-05-19
overlay: Resolve phandle fixups in child nodes

arthokal:
**Issue**:
When user overlay files (passed via `-i`) contain child nodes with phandle references, those references are silently dropped in the generated pl.dtsi

For example, `updated_zocl.dtsi` contains:
```
&amba_pl {                                                                      
        zyxclmm_drm {
                compatible = "xlnx,zocl-versal";
                xlnx,memory-region = <&cma_reserved>,<&ddr_ch1>,<&ddr_ch2>,<&ddr_ch3>;
        };
};
```
The generated overlay has `zyxclmm_drm` with `xlnx,memory-region` missing.

**Root Cause**:
`_unwrap_overlay_tree()` in `lopper/__init__.py` resolves `0xffffffff` phandle placeholders (left by dtc plugin compilation) only for direct properties of the `__overlay__` node.
Child node properties (like those in `zyxclmm_drm`) are deep-copied as-is with unresolved `0xffffffff` values, which are silently dropped at write time.

**Fix**:
Instead of iterating properties and searching for matching fixups (inside-out), iterate `label_to_fixups` and patch nodes directly via `ov_tree[path] (outside-in). This resolves phandles across `__overlay__` and all its children in one pass before any copying happens.

**Example Lopper command**:
```
LOPPER_DTC_FLAGS="-b 0 -@" ./lopper.py -O outdir -f --enhanced -i /proj/ssw_xhd/drivers/onkarh/to_aravind/updated_zocl.dtsi /proj/ssw_xhd/drivers/onkarh/to_aravind/apu_linux-chosen.dts system-top-no-pl.dts -- xlnx_overlay_pl_dt psv_cortexa72_0 full --firmware-name=gmio_async_xrt.pdi
```


arthokal:
@onkarharsh,
Please review.

zeddii:
The current resolution architecture was on purpose and should only be done on write time.

I don't have access to my full setup to test this, as I'm on vacation this week.

Does this change preserve the phandle resolution at write time ? I can't tell from looking at the patch.

Could you also attach some sample inputs and expected outputs ? So I can reproduce and test the issue myself.

arthokal:
Hi @zeddii ,

Sorry for the delay in response.

The original code already resolves phandle fixups at unwrap time. This patch only extends that same resolution to child nodes, which were being missed.

I emailed you the artifacts to reproduce the issue. Please check.

CC: @onkarharsh

zeddii:
Thanks for identifying the real problem here — child node phandle placeholders were not being resolved, and the root cause analysis is correct.

After reviewing the PR the concern with extending resolution at unwrap/registration time still stands: the base tree can be mutated by assists and lops *after* registration but *before* `overlay_tree()` is called (node renames, phandle renumbering, etc.), so any resolution done at registration time is against a potentially stale snapshot.

I've taken a different approach that handles child nodes and addresses the architectural concern at the same time, landed as two commits on master:

**`f484b0da` — overlay: defer phandle resolution to build time via _resolve_overlay_fixups**

`_unwrap_overlay_tree()` now preserves `0xffffffff` placeholders as-is and rewrites the dtc fragment paths to real abs_paths, but does no phandle patching. All resolution is deferred to `_resolve_overlay_fixups()`, called from `_build_overlay_tree()` after the full merged result tree is assembled. This means resolution always runs against the final tree state — after all assists have had a chance to modify it — rather than the base tree snapshot at registration time. Child nodes at any depth are covered because `_resolve_overlay_fixups()` looks up nodes by path in the merged result tree regardless of depth.

**`180c06d2` — overlay: store fixup refs as (label, rel_path) tuples, not abs_paths**

A follow-on observation: storing rewritten abs_paths in the fixup metadata is redundant and fragile — if an assist renames the fragment target node (e.g. `/amba_pl` → `/axi/amba_pl`) between registration and `overlay_tree()` build time, the stored path becomes stale. The fixup entries now store `(fragment_label, relative_path, prop_name, byte_offset)` tuples instead. At build time `_resolve_overlay_fixups()` looks up the fragment label in the result tree to get its current abs_path, then appends the relative child path — transparent to any renames.

The net result is the same as this PR for the child node case, but the resolution is done at the correct point in the pipeline.

Closing this PR in favour of the above commits. Thank you for the report.

zeddii:
Closing. Let me know if your tests aren't passing with these changes (they are for me), and we can revisit.

PR #767, opened 2026-05-25
lopper:assists:baremetal_xparameters_xlnx: Refactor DDRMC5 property e…

sathishkumar-amd:
…xtraction to use target node instead of axi_noc2 nodes

Read the DDR5 memory controller properties (DDRMC5_I2C_MASTER and DDRMC5_DEBUG_ELF) from the root node instead of the AXI NoC2 memory node


sathishkumar-amd:
Hi @onkarharsh,

Please review this

onkarharsh (approved):
Looks good to me

PR #768, opened 2026-05-28
lopper: assists: zephyr: Drop bootph-all property from generated doma…

kedareswararao:
…in DTS

Zephyr enforces strict property checks on DTS nodes and does not recognize bootph-all. Its presence causes build-time errors when the generated domain DTS is consumed by Zephyr.

Strip the bootph-all property from clock nodes during domain DTS generation, consistent with how reg and clock-output-names are already excluded.


kedareswararao:
@zeddii : Could you please merge this fix currently zephyr mb-v builds got broken with latest sdt this pull request fixes the same.

PR #771, opened 2026-06-15
lopper:assists:petalinuxconfig_xlnx: remove redundant iterations

parthibank1:
petalinuxconfig_xlnx emitted memory controllers such as lpddrmc as plain slave entries, but did not emit the corresponding DDR memory node with device_type, baseaddr, and highaddr metadata.

For MB-V SCU200, petalinux_config.yaml contained lpddrmc_0 with only ip_name: lpddrmc, while the DDR memory node was missing:
      lpddrmc_0_ddr_memory:
          device_type: memory
          ip_name: lpddrmc
          baseaddr: '0x80000000'
          highaddr: '0xffffffff'

Since kconfig_syshw.py creates memory choices from entries with memory metadata, DDR was not available as a selectable memory bank. Kconfig fell back to manual memory at 0x0, which propagated into the generated machine
configuration as DDR_BASEADDR=0x0 and RISCV_SBI_FW_TEXT_START=0x0.

This is invalid for OpenSBI because FW_TEXT_START points to BRAM/low memory instead of DDR, causing the OpenSBI link to fail.

Fix this by fetching memory ranges once when memory device types exist and emitting the resolved memory nodes into petalinux_config.yaml using their DT labels. This makes DDR available to kconfig_syshw.py and lets MB-V SCU200
select DDR at 0x80000000 instead of falling back to 0x0.

Also avoid redundant work in the same path by calling get_memranges() once instead of once per memory schema entry, and by processing mapped_nodelist once through a device-type map while preserving unmatched-device fallback behavior.

For machine configurations that already emitted complete memory metadata, the generated output remains equivalent. The output changes only for affected machine configurations where valid memory nodes were previously dropped, such
as lpddrmc_0_ddr_memory on MB-V SCU200.

### DIFF
--- a/lopper/assists/petalinuxconfig_xlnx.py
+++ b/lopper/assists/petalinuxconfig_xlnx.py
@@ -72,16 +72,29 @@ def xlnx_generate_petalinux_config(tgt_node, sdt, options):
     4. For uart device_type add default baud rate 
     5. ???
     """
-    tmp_dict = {}
+    tmp_dict = {'slaves': {}}  # Initialize slaves dict to prevent KeyError
     proc_name = ""
     device_type_dict = {}
+
+    if not os.path.isfile(yaml_file):
+        print(f"Error: YAML file not found: {yaml_file}")
+        return False
+
     with open(yaml_file, 'r') as stream:
         schema = yaml.safe_load(stream)
         res_list = list(schema.keys())
-        tmp_node_list = []
         device = sdt.tree['/'].propval('device_id')
         if device != ['']:
             device_type_dict['device_id'] = device[0]
+
+        # Check if memory device type exists in schema before fetching memory ranges
+        has_memory_type = any(re.search("memory", schema[res]['device_type']) for res in res_list)
+        if has_memory_type:
+            mem_ranges, label_names = get_memranges(tgt_node, sdt, options)
+            memnode_list = sdt.tree.nodes('/memory@.*')
+
+        # Build device type mapping for non-processor/non-memory devices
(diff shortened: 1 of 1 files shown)


kedareswararao:
> https://jira.xilinx.com/browse/CR-1269946
> 
> * Pre-fetch memory ranges and node lists before the schema loop to avoid redundant get_memranges() calls when multiple memory device types exist
> * Eliminate double iteration of mapped_nodelist by building a device_type_map first, then iterating nodes once with both matching and fallback logic
> * Only fetch memory data when schema contains memory device types
> * Move ipname property access to where it's actually needed

don't mention JIRA numbers/links in the github pull requests

sathishkumar-amd (approved):
Looks good to me

Rajukumar45 (approved):
Looks good to me

zeddii:
The commit message for the change only describes what it is doing.

It doesn't describe why it is being done.

Were there performance issues ? Was there a bug/issue ? or is this just a cosmetic / tidy up of the code ?

The commit message should also confirm that the output is the same after the change, as before.

parthibank1:
@zeddii Updated the PR description.

zeddii:
Thanks, that's the type of description I was looking for.

But it will be lost if only captured in the github pull request. 

Can you amend the commit and add the details ? I can do it on your behalf if you don't have time.

parthibank1:
@zeddii Yes Please, that would be helpful.

zeddii:
I've re-worded the commit, and it is now in master

PR #775, opened 2026-06-19
lopper: assists: xlnx_overlay_pl_dt: Handle "segmented" config in FPGA node creation

arthokal:
**Root Cause**:
The "segmented" config was accepted by validate_and_parse_options() but had no branch in create_fpga_node(), so the default /amba_pl extraction path fell into the else and exited with "Invalid configuration: segmented" failing overlay generation for segmented designs

**Fix**:
The legacy xlnx_overlay_dt assist treated "segmented" the same as "full": it retained the firmware-name property and emitted the overlay normally. Restore that behavior by handling "segmented" alongside "full" in create_fpga_node(), so the FPGA node keeps firmware-name for all platforms.

The AFI/clocking placement in move_nodes_to_fpga() already keys off config != "external-fpga-config", so segmented is handled there without change.

Docstrings and comments are updated to list segmented among the valid configs.

### DIFF
--- a/lopper/assists/xlnx_overlay_pl_dt.py
+++ b/lopper/assists/xlnx_overlay_pl_dt.py
@@ -420,7 +420,7 @@ def create_fpga_node(platform, config, firmware_name, zynq_platforms, versal_pla
 
     Args:
         platform: Target platform (e.g., cortexa53-zynqmp, cortexa72-versal)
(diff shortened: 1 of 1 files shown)


arthokal:
Hi @onkarharsh ,
Please review.

onkarharsh (approved):
Looks good to me

PR #776, opened 2026-06-22
lopper: assists: gen_domain_dts: Allow passing extra zephyr comp file…

KasperSloth99:
… through env

I ran into an issue of not being able to add my own zephyr_supported_comp.yaml file, without having to have my own fork and edit the upstream file. Instead it is much nicer to be able to be able to set the path to your own compatible schema, and have it be part of your own repository.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -11,6 +11,7 @@ import yaml
 import sys
 import os
 import glob
+import argparse
 import lopper
 import re
 from lopper.tree import LopperProp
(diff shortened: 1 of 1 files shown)


KasperSloth99:
Hi @onkarharsh please review

zeddii:
environment variable only options are clunky. You can keep the environment variable processing, but also passing it as a command line options makes it much more discoverable and obvious in logs and command lines.

KasperSloth99:
@zeddii I just followed how it is done with f.ex LOPPER_DTC_FLAGS. If you prefer i can do it as a CLI option?

zeddii:
LOPPER_DTC_FLAGS is one of the few environment variables (and for what it is used for, it shouldn't be used like it is anymore, there's a command line equivalent). 

For something down in the pipeline like an assist, having the option near the assist on the command line helps usability and understandability.

KasperSloth99:
@zeddii I changed it to be a cli argument instead.

zeddii:
This looks good to me. Still waiting on the other review before merging.

kedareswararao (approved):
Changes looks fine to me

PR #777, opened 2026-06-24
assists: xlnx_overlay_pl_dt: auto-discover *.dtso files from SDT input directory

arthokal:
Instead of requiring -i flags for every lopper invocation to include custom overlay properties for PS nodes, auto-discover *.dtso files co-located with system-top.dts and register them through the same compilation path used by the -i flag.

discover_overlay_files() scans the directory containing the input SDT file for *.dtso files, skips any already registered via -i to avoid duplicates, and calls _compile_overlay_subtrees() to compile and register them. fragment_add_for_refs() then emits them as &label overlay fragments in pl.dtsi automatically.

### DIFF
--- a/lopper/assists/xlnx_overlay_pl_dt.py
+++ b/lopper/assists/xlnx_overlay_pl_dt.py
@@ -769,6 +769,43 @@ Output:
     - pl.dtso: Copy of pl.dtsi with .dtso extension
     Note: Modified system device tree (with /amba_pl node removed) is written by lopper itself
 """
+
+def discover_overlay_files(sdt):
+    """Auto-discover *.dtso overlay files co-located with system-top.dts.
+
+    Scans the directory containing the input SDT file for files matching
+    the *.dtso extension and registers them via the same compilation path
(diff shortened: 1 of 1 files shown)


arthokal:
@onkarharsh , please review.

onkarharsh (approved):
Looks good to me

zeddii:
I'd strongly suggest that you have a way to disable this.  In case the user has files that they actually don't want processed (they could be large, they could have errors, etc).

So just a simple flag to the assist to opt-out (or make an opt-in flag, you choice) would give an escape hatch.

arthokal:
Hi @zeddii ,

Added --exclude-overlays option to selectively skip .dtso files from auto-discovery. It accepts a comma-separated list of filenames with glob support:

skip a specific file:
--exclude-overlays=debug.dtso
skip multiple files:
--exclude-overlays=a.dtso,b.dtso
skip all (disable auto-discovery):
--exclude-overlays=*.dtso

Please review.

zeddii:
looks good to me. this is merged.

PR #778, opened 2026-06-25
assists: bmcmake_metadata_xlnx: fix type mismatch

venodela:
Generated lwip xtopology_g.c used NULL to zero-terminate the xtopology array, triggering -Wint-conversion because the field type is long unsigned int, not a pointer. Use integer literal 0 instead.


sathishkumar-amd (approved):
Looks good to me

venodela:
cc : @kedareswararao @HariniKatakamX

PR #780, opened 2026-06-29
lopper:lops:lop-microblaze-riscv: Add -mcmodel=medany compiler flag f…

sathishkumar-amd:
…or MicroBlaze RISC-V

MicroBlaze RISC-V designs that map memory at or above 0x80000000 (2 GiB) require the -mcmodel=medany compiler flag. The default medlow code model cannot address such regions and leads to build/runtime issues.


sathishkumar-amd:
Hi @onkarharsh,
Please review this

onkarharsh:
Please use propval() instead of try/except block

sathishkumar-amd:
> Please use propval() instead of try/except block

Updated the code to use propval()

onkarharsh (approved):
Looks good to me

PR #782, opened 2026-07-03
lopper: assists: Remove xlnx,bus-width for ADMA

abinJoseph25k07:
Remove xlnx,bus-width from the yaml and gen_domain_dts file


abinJoseph25k07:
cc: @kedareswararao

kedareswararao (commented):
Please update the commit description with the reason why we are doing this change

kedareswararao (approved):

PR #785, opened 2026-07-09
lopper:assists:bmcmake_metadata_xlnx: Set CPM6_SUPPORTED CMake variab…

sathishkumar-amd:
…le from SDT

Detect the xlnx,cpm6_supported property on the matched CPU node in the system device tree and emit CPM6_SUPPORTED=ON in the generated CMake metadata when CPM6 is present in the design. This bridges the SDT CPM6 detection to the embeddedsw xilpm build configuration, enabling automatic inclusion of GPIO proc handler support for CPM6-based Versal designs without manual BSP configuration.


onkarharsh (approved):
Looks good to me

PR #786, opened 2026-07-09
lopper:assists: Add RV32E (embedded) compiler/ABI support for MicroBlaze-V.

sushil-dev-amd:
Propagate the xlnx,use-embedded CPU property (set by SDTGen for MicroBlaze-V cores with C_USE_EMBEDDED) through to the generated build flags:

- lop-microblaze-riscv.dts: emit rv32e/-mabi=ilp32e in cflags.yaml instead of rv32i/ilp32 when the core is configured as embedded, and apply the ABI flag to linkflags as well as cflags so BSP link-time multilib selection matches. Guard against the invalid use-embedded + hardware-FPU combination (RV32E has no standard hard-float ABI) and warn if use-embedded is set alongside a 64-bit data size (RV32E is 32-bit only).
- gen_domain_dts.py: select RISCV_ISA_RV32E instead of RISCV_ISA_RV32I in the generated Zephyr SoC and board Kconfig when the parsed ISA string indicates an embedded (rv32e) core.

Signed-off-by: sushilkumar.singh@amd.com

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -1451,12 +1451,7 @@ def generate_board_kconfig_defconfig(isa_string, cpu_node, intc_node, num_interr
     # Track added extensions to avoid duplicates
     added_extensions = set()
 
-    # Base ISA configurations - always add RV32I for MicroBlaze RISC-V
-    content += "config RISCV_ISA_RV32I\n"
-    content += "\tdefault y\n\n"
-    added_extensions.add('RISCV_ISA_RV32I')  # Track it to avoid duplicates
-
-    # Process ISA string to generate configs in the exact order: RV32I, M, A, C, ZICSR, ZIFENCEI
+    # Process ISA string to generate configs in the exact order: RV32I/RV32E, M, A, C, ZICSR, ZIFENCEI
     if isa_string:
         # Define the exact order matching the existing board defconfig
         ordered_extensions = [
@@ -1469,6 +1464,13 @@ def generate_board_kconfig_defconfig(isa_string, cpu_node, intc_node, num_interr
 
         # Process base ISA part for M, A, C detection
         isa_base = isa_string.split('_')[0]  # Same as SOC level
+
+        # Base ISA configuration - RV32E for embedded (reduced register set) MB-V
+        # cores (xlnx,use-embedded), RV32I otherwise.
+        base_isa_config = 'RISCV_ISA_RV32E' if isa_base.startswith('rv32e') else 'RISCV_ISA_RV32I'
+        content += f"config {base_isa_config}\n"
+        content += "\tdefault y\n\n"
+        added_extensions.add(base_isa_config)  # Track it to avoid duplicates
 
         # Process each extension in the defined order
         for key, config_name in ordered_extensions:
@@ -1512,6 +1514,12 @@ def generate_board_kconfig_defconfig(isa_string, cpu_node, intc_node, num_interr
                 content += f"config {config_name}\n"
                 content += "\tdefault y\n\n"
                 added_extensions.add(config_name)
+    else:
+        # No ISA string available - fall back to the default RV32I base,
+        # matching the prior (pre-RV32E) unconditional behavior.
+        content += "config RISCV_ISA_RV32I\n"
+        content += "\tdefault y\n\n"
+        added_extensions.add('RISCV_ISA_RV32I')
 
     content += "endif\n"
     return content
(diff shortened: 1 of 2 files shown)


sushil-dev-amd:
@sathishkumar-amd, @sivadur and @kedareswararao, please review the downstream compilation flags support to enable rv32e compilation for 32-bit arch using use_embedded flag (C_HAS_EMBEDDED parameter) from the DT. 
Note:  use_embedded and use_fpu cannot be 1 at the same time, since there exists no fpu for rv32e support.

lopper/assists/gen_domain_dts.py
```diff
@@ -1451,10 +1451,13 @@ def generate_board_kconfig_defconfig(isa_string, cpu_node, intc_node, num_interr
     # Track added extensions to avoid duplicates
     added_extensions = set()
 
-    # Base ISA configurations - always add RV32I for MicroBlaze RISC-V
-    content += "config RISCV_ISA_RV32I\n"
+    # Base ISA configuration - RV32E for embedded (reduced register set) MB-V
+    # cores (xlnx,use-embedded), RV32I otherwise.
+    isa_base_check = isa_string.split('_')[0] if isa_string else ''
```
kedareswararao: can you add check to make sure isa_string variable is not empty like if isa_string: and keep your code inside it?

lopper/lops/lop-microblaze-riscv.dts
```diff
@@ -152,10 +158,21 @@
                                        archflags.append('_zicbom')
                                        bsp_archflags.append('_zicbom')
 
+                                   use_embedded = n.propval('xlnx,use-embedded')
+                                   is_embedded = use_embedded != [''] and use_embedded[0] == 1 and n['xlnx,data-size'].value[0] != 64
+
+                                   if is_embedded and n['xlnx,use-fpu'].value[0] != 0:
+                                       raise Exception('Invalid MicroBlaze-V configuration at {}: xlnx,use-embedded=1 (RV32E/ilp32e) is not compatible with a hardware FPU (xlnx,use-fpu={}). The RV32E ABI has no standard hard-float variant; disable the FPU or the embedded (RV32E) option.'.format(n.abs_path, n['xlnx,use-fpu'].value[0]))
```
kedareswararao: so this configuration is allowed at Vivado level and not valid? should we be asking  the IP owner to update it such away that vivado GUI to not to allow this configuration?

lopper/assists/gen_domain_dts.py
sushil-dev-amd: okay will add it inside the if block below.

lopper/lops/lop-microblaze-riscv.dts
sushil-dev-amd: Yes ideally, it should be blocked at hardware IP design level, and lopper layer should not decide on this. But not sure who to reach out for implementing this configuration logic in hardware level.

sushil-dev-amd:
@kedareswararao, @sathishkumar-amd and @sivadur, please review the latest changes.

lopper/assists/gen_domain_dts.py
```diff
@@ -1470,6 +1465,13 @@ def generate_board_kconfig_defconfig(isa_string, cpu_node, intc_node, num_interr
         # Process base ISA part for M, A, C detection
         isa_base = isa_string.split('_')[0]  # Same as SOC level
 
+        # Base ISA configuration - RV32E for embedded (reduced register set) MB-V
+        # cores (xlnx,use-embedded), RV32I otherwise.
```
kedareswararao, line 1469: can you cross check this configuration is supported in zephyr SDK toolchain if not supported please drop changes for zephyr rest of the changes looks fine please confirm this will approve

zeddii:
Bumping this — @sushil-dev-amd, was there any follow-up on @kedareswararao's question about whether this configuration is supported in the Zephyr SDK toolchain? Per the review, if it isn't supported the Zephyr changes should be dropped (the rest of the change looks fine). We won't merge this until that's cross-checked and we have the all-clear from the reviewer.

lopper/assists/gen_domain_dts.py
sushil-dev-amd: Hi as per discussion will try to use the toolchain zephyr internally and update on this.

sushil-dev-amd: yes, tested this and see the config RISCV_ISA_RV32E being generated in zephyr/boards/amd/mbv32/Kconfig.defconfig with the design supporting C_USE_EMBEDDED IP config.

kedareswararao: okay then i am fine with these changes approving

kedareswararao (approved):

PR #790, opened 2026-07-21
lopper: generate Zephyr board.overlay from SDT folder

dbingi-amd:
SDTGEN now copies Zephyr board fragments into the SDT output folder as sidecar .dtsi files ({board}_zephyr.dtsi and optional user input via -user_zephyr_dts), instead of passing a single board .dts file to lopper.
This PR adds SDT-folder-based discovery and board.overlay generation in lopper, and replaces the old single-file overlay handling in gen_domain_dts. A follow-up commit adds pytest unit tests to validate discovery rules and all three overlay flows
• Add discover_zephyr_board_files() to find {board}_zephyr.dtsi and user .dtsi files.
• Treat user DTSI as any other .dtsi not #included from system-top.dts.
• Add generate_board_overlay_from_sdt() for board-only, board+user, and user-only flows.
• Filter board overlay against the SDT tree; append user overlay unchanged when present.
• Update gen_domain_dts.py to call the new helper instead of reading one board file.
• Add tests/test_zephyr_board_overlay.py for discovery and overlay generation unit tests

### DIFF
--- a/lopper/assists/zephyr_board_dt.py
+++ b/lopper/assists/zephyr_board_dt.py
@@ -7,302 +7,297 @@
 # * SPDX-License-Identifier: BSD-3-Clause
 # */
 
-import yaml
-import sys
-import os
+import copy
 import glob
-import lopper
+import os
 import re
-from lopper.tree import LopperProp
-from lopper.tree import LopperNode
+import sys
+import tempfile
+
+from lopper import Lopper, LopperSDT, compile_overlay_standalone, _unwrap_overlay_tree
+from lopper.tree import _merge_node_into_tree, _resolve_overlay_fixups
 
 sys.path.append(os.path.dirname(__file__))
 
-import sys
-import os
-import re
+OVERLAY_REF_RE = re.compile(r"&\w+\s*{")
+ZEPHYR_BOARD_DTSI_SUFFIX = "_zephyr.dtsi"
+SDT_DTSI_GLOB = "*.dtsi"
+SDT_TOP_DTS = "system-top.dts"
+SDT_BOARD_PROP_RE = re.compile(r'board\s*=\s*"([^"]+)"')
+SDT_INCLUDE_RE = re.compile(r'#include\s+"([^"]+)"')
 
 
-def process_overlay_with_lopper_api(overlay_content, main_tree):
-    """
-    Process overlay content using official lopper tree API methods
-    
-    Args:
-        overlay_content (str): The overlay file content
-        main_tree: The LopperTree object (sdt.tree)
-    
-    Returns:
-        str: Cleaned overlay content
-    """
-    
-    # Extract all overlay references (&reference)
-    overlay_refs = set(re.findall(r'&(\w+)', overlay_content))
-    
-    # Extract internal labels (locally defined in the overlay)
-    internal_labels = set(re.findall(r'(\w+):\s*[\w@-]+\s*{', overlay_content))
-    
-    #print(f"[INFO] Found overlay references: {sorted(list(overlay_refs))}")
-    #print(f"[INFO] Found internal labels: {sorted(list(internal_labels))}")
-    
-    # Check which references exist using lopper tree API
-    valid_refs = []
-    invalid_refs = []
-    
-    for ref in overlay_refs:
-        found_in_system = False
-        found_internally = ref in internal_labels
-        
-        #print(f"[INFO] Checking reference '{ref}' (internal: {found_internally})")
-        
-        if not found_internally:
-            # Use lopper tree API calls to check if reference exists
-            found_in_system = check_reference_with_lopper_api(main_tree, ref)
-        
-        if found_in_system or found_internally:
-            valid_refs.append(ref)
-        else:
-            invalid_refs.append(ref)
-    
-    # Process overlay content to remove invalid sections
-    if not invalid_refs:
-        #print("[INFO] All references are valid. Applying final cleanup only.")
-        return apply_final_cleanup(overlay_content)
-    
-    return clean_overlay_content(overlay_content, valid_refs, invalid_refs, internal_labels)
+def _include_paths(sdt_folder, sdt):
+    paths = [p for p in (sdt_folder, getattr(sdt, "outdir", None)) if p and os.path.isdir(p)]
+    return " ".join(paths)
 
 
-def check_reference_with_lopper_api(tree, ref_name):
-    """
-    Check if a reference exists using official lopper tree API methods
-    Fixed version with accurate debug output
-    
-    Args:
-        tree: The LopperTree object
-        ref_name: The reference name to search for
-    
-    Returns:
-        bool: True if reference is found, False otherwise
-    """
-    
-    # Method 1: Check aliases using tree['path'] access
-    try:
-        aliases_node = tree['/aliases']
-        if aliases_node:
-            # Check if this reference is an alias
-            try:
-                alias_prop = aliases_node[ref_name]
-                if alias_prop:
-                    #print(f"[DEBUG] ✓ Found alias '{ref_name}' = '{alias_prop.value}' in tree['/aliases']")
-                    return True
-            except:
-                pass
-        #print(f"[DEBUG] ✗ No alias '{ref_name}' found in tree['/aliases']")
-    except Exception as e:
-        #print(f"[DEBUG] ✗ Aliases check failed for '{ref_name}': {e}")
-        pass
-    
-    # Method 2: Check symbols using tree['path'] access
-    try:
-        symbols_node = tree['/__symbols__']
-        if symbols_node:
-            try:
-                prop_dict = symbols_node.__props__
-                symbol_exist = [label for label,node_abs in prop_dict.items() if label == ref_name]
-                if symbol_exist:
-                    #print(f"[DEBUG] ✓ Found symbol '{ref_name}' -> '{symbol_prop.value}' in tree['/__symbols__']")
-                    return True
-            except:
-                pass
-        #print(f"[DEBUG] ✗ No symbol '{ref_name}' found in tree['/__symbols__']")
-    except Exception as e:
-        #print(f"[DEBUG] ✗ Symbols check failed for '{ref_name}': {e}")
-        pass
-    
-    # Method 3: Use subnodes to traverse the tree
-    try:
-        root_node = tree['/']
-        if root_node:
-            # Get direct children first
-            subnodes_list = tree.subnodes(root_node)
-            for node in subnodes_list:
-                # Check node label
-                if hasattr(node, 'label') and node.label == ref_name:
-                    #print(f"[DEBUG] ✓ Found node with label '{ref_name}' at: {node.abs_path}")
-                    return True
-                
-                # Check node name
-                if hasattr(node, 'name'):
-                    node_name = node.name
-                    # Match exact or with address suffix
-                    if node_name == ref_name or node_name.startswith(f"{ref_name}@"):
-                        #print(f"[DEBUG] ✓ Found node '{ref_name}' by name: {node_name} at {node.abs_path}")
-                        return True
-                
-                # Recursively check subnodes
-                if _check_subnodes_recursive(tree, node, ref_name):
-                    return True
-    except Exception as e:
-        #print(f"[DEBUG] ✗ Subnodes traversal failed for '{ref_name}': {e}")
-        pass
-    
-    #print(f"[DEBUG] ✗ Reference '{ref_name}' not found using any lopper tree API method")
-    return False
-
- 
-def apply_final_cleanup(content):
-    """
-    Apply final formatting cleanup to remove artifacts
-    """
-    # Remove multiple consecutive empty lines
-    content = re.sub(r'\n\s*\n\s*\n+', '\n\n', content)
-    
-    # Remove empty alias blocks
-    content = re.sub(r'aliases\s*{\s*}\s*;?', '', content)
-    
-    # Remove orphaned semicolons on their own lines
-    content = re.sub(r'^\s*;\s*$', '', content, flags=re.MULTILINE)
-    
-    # Clean up spacing around braces
-    content = re.sub(r'}\s*;\s*\n', '};\n', content)
-    
-    # Remove trailing whitespace from lines
-    content = re.sub(r'[ \t]+$', '', content, flags=re.MULTILINE)
-    
-    # Ensure single newline at end
-    content = content.rstrip() + '\n'
-    
-    # One more pass to clean up multiple newlines that might have been created
-    content = re.sub(r'\n\s*\n\s*\n+', '\n\n', content)
-    
-    return content
+def _work_dir(sdt):
+    tmpdir = getattr(sdt, "tmpdir", None) if sdt else None
+    if tmpdir and os.path.isdir(tmpdir):
+        return tmpdir
+    outdir = getattr(sdt, "outdir", None) if sdt else None
+    return tempfile.mkdtemp(prefix="zephyr_board_", dir=outdir or None)
 
 
-def clean_overlay_content(overlay_content, valid_refs, invalid_refs, internal_labels):
-    """
-    Clean the overlay content by removing invalid sections
-    """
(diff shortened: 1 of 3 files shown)


dbingi-amd:
@kedareswararao: please review the changes

lopper/assists/zephyr_board_dt.py
```diff
@@ -306,3 +306,168 @@ def _check_subnodes_recursive(tree, node, ref_name, max_depth=5, current_depth=0
...
+
+def generate_board_overlay_from_sdt(sdt, options):
+    """
+    Generate board.overlay during Zephyr DTS generation.
+
+    Scenario 1: board DTS only -> validate/filter against SDT tree
+    Scenario 2: board + user DTS -> filter board, append user unchanged
+    Scenario 3: user DTS only -> copy user content unchanged
+    """
+    sdt_folder = resolve_sdt_folder(options, sdt)
```
kedareswararao, line 280: Here we should support legacy use case as well where users pass the board dts instead of sdt folder

lopper/assists/zephyr_board_dt.py
```diff
@@ -306,3 +306,168 @@ def _check_subnodes_recursive(tree, node, ref_name, max_depth=5, current_depth=0
...
+        filtered_board = process_overlay_with_lopper_api(
+            _read_text_file(board_dtsi), sdt.tree
+        ).rstrip()
+        if filtered_board:
+            overlay_parts.append(filtered_board)
+        overlay_parts.append(_read_text_file(user_zephyr_dtsi).rstrip())
+        print(f"[INFO] Generated board.overlay from filtered '{os.path.basename(board_dtsi)}' and user '{os.path.basename(user_zephyr_dtsi)}'.")
+    elif board_dtsi:
+        overlay_parts.append(
+            process_overlay_with_lopper_api(
```
kedareswararao: This process_overlay_with_lopper_api() uses pure python calls recently Bruce added support for handling this kind of use cases in lopper frontend can you see whether we can remove the python API's and replace it with lopper API's

lopper/assists/zephyr_board_dt.py
dbingi-amd: Acknowledged. I will add backward compatibility

lopper/assists/zephyr_board_dt.py
dbingi-amd: I reviewed the Lopper overlay frontend APIs, but they are designed for true Device Tree overlays that are merged into the SDT tree. In contrast, our Zephyr board use cases rely on .dtsi include fragments containing both root-level blocks and node-specific sections. The reference validation logic already uses the Lopper tree APIs, and the remaining text-filtering functionality does not have a direct equivalent in the overlay frontend APIs. Therefore, we chose to retain this helper implementation.

kedareswararao (approved):
Changes looks fine to me

zeddii:
Thanks for the work here. Before this lands I'd like to redirect the overlay handling, because the current approach re-implements a large amount of what lopper core (and dtc) already do, with text processing that will be fragile to maintain.

**What lopper core already does with `.dtsi` fragments**

A `.dtsi` is an include-and-compile artifact. When it's included and compiled, core already gives you everything this PR does by hand:
- `&label` references are resolved by lopper's symbol resolution against the base tree;
- properties whose phandle references don't resolve are dropped on write (strict mode, the default);
- a reference to a label that genuinely doesn't exist is a hard compile error — which is what we want surfaced, not silently deleted.

In fact a `.dtsi` handed to lopper as an input is *already* concatenated with the base and compiled — `setup()` routes non-`/plugin/` `.dtsi` into the base compile unit. So this isn't new machinery; the assist just needs to do the same thing internally.

**Where this PR duplicates that**

`process_overlay_with_lopper_api()` and its helpers reimplement the above in Python/regex:
- `re.findall(r'&(\w+)', …)` / `re.findall(r'(\w+):\s*…{', …)` extract references and labels by regex instead of parsing;
- `check_reference_with_lopper_api()` / `_check_subnodes_recursive()` hand-walk `/aliases`, `/__symbols__` and recurse the tree to test whether a symbol exists — that is core symbol resolution, re-derived;
- `clean_overlay_content()` brace-counts over raw text to delete `&ref { … }` blocks;
- `apply_final_cleanup()` regex-scrubs the formatting.

That's several hundred lines duplicating the compiler and lopper's resolver, and text/brace/regex handling of DTS will break on the first fragment that doesn't match the assumed formatting. The root problem is that these `.dtsi` are processed *externally* when they should be included and compiled.

**Path forward — step 1: include + compile**

Keep taking the fragment as an assist argument — the invocation doesn't need to change. The fix is what the assist *does* with it: hand it to the compiler instead of parsing it. Compose the base + fragment and compile (note `dt_compile()` doesn't consume `i_files` — includes are handled by concatenation, the same way `setup()` seeds base `.dtsi` inputs):

```python
import os, shutil
from lopper import Lopper
from lopper.tree import LopperTree

system_top = os.path.join(sdt_folder, "system-top.dts")

# compose base + fragment into one source, then compile it
combined_src = os.path.join(sdt.outdir, "_board_combined.dts")
with open(combined_src, "wb") as out:
    for src in (system_top, board_dtsi):
        with open(src, "rb") as f:
            shutil.copyfileobj(f, out)

compiled, _ = Lopper.dt_compile(combined_src, "", "", True,
                                sdt.outdir, sdt.save_temps)

combined = LopperTree()
combined.load(Lopper.export(Lopper.dt_to_fdt(compiled)))
# combined = base + fragment, fully resolved by core:
#   &label refs resolved, unresolved phandle refs dropped on write,
#   an undefined label is a hard compile error. No regex, no
#   __symbols__ walking, no brace-counting.
```

(Equivalently, register the fragment into the SDT's input files so `setup()` concatenates it — whatever fits the assist structure.) Either way this needs nothing new from lopper core.

**Step 2 depends on one question about the Zephyr side**

What consumes `board.overlay` — does the Zephyr build require a *separate* overlay file, or can it take a complete generated board device tree?
- If it can take a full board DT, we're done at step 1: emit `combined` and there's no overlay to filter at all.
- If a separate overlay file *is* required (Zephyr owns a base board DT that must be *augmented*, not replaced), the right way to produce it is a tree-delta — compare `combined` against the base and emit the difference as an overlay fragment. That belongs as a general lopper capability (advanced tree delta / diff with an overlay output format), not text filtering in one assist. **If that's the requirement, I'll take care of that piece.**

Could you confirm the Zephyr consumption? That decides between "emit the combined tree" and "emit a computed delta," and either way the hand-rolled parsing should come out.

(+1 to keeping the legacy single-file path working.)

kedareswararao:
@zeddii : Thanks for the suggestion combined dt also works for zephyr consumption 
@dbingi-amd : give a try and let us know your observations

dbingi-amd:
@zeddii,
Thanks for taking the time to review this.

Step 1: I switched to the lopper compile/include path instead of regex.
- I removed process_overlay_with_lopper_api() and all regex/brace-based overlay text filtering. Merge now relies on the lopper core APIs:
  • &label { } content (for example, kcu105_zephyr.dtsi): compile_overlay_standalone → _unwrap_overlay_tree → _merge_node_into_tree → _resolve_overlay_fixups, plus root /aliases and /chosen merge with fragment path rewriting.
  • Root-only / { ... } content (no & refs): domain base + fragment concat through Lopper.dt_compile() / LopperSDT.setup() — the same pattern used for lopper include handling.

- A plain concat + dt_compile path is not enough for a full board .dtsi with & refs, because optional or missing labels (for example, axi_ethernet_0 on kcu105) would cause the entire compile to fail; the plugin path merges the matching fragments and skips the rest.

Step 2: I split out overlay vs. combined board DT.
- Per @kedareswararao's direction, I implemented a combined board DT. During gen_domain_dts, the Board/user Zephyr DTSI is merged into sdt.tree, and no board.overlay is generated. Zephyr uses the single merged board .dts (for example, mbv32.dts).

@kedareswararao:
- I tested end to end and verified the combined DT for zephyr.
Observation: the reference board .dtsi may include optional IP that is not present in every XSA; unmatched & fragments are skipped with a warning, while matched content and aliases are merged. We can switch to a hard failure on missing labels if preferred.

zeddii:
Thanks -- good progress, and glad the combined DT works for the Zephyr side.

On the missing-label question: **default should be an error.** dtc won't build a plain `.dtsi` with an unresolved label (I checked -- even `-f` is fatal: *"Label or path ... not found"*), so allowing it silently means the `.dtsi` isn't really a `.dtsi` anymore. If a file legitimately needs to reference optional IP that may be absent in a given XSA, that is **overlay** content -- dtc supports undefined labels there via `__fixups__`, and that's the mechanism you're already using through `compile_overlay_standalone` / `_unwrap_overlay_tree`. So those reference board files should be treated as overlays, not include fragments.

If we do want lopper to tolerate an unresolved label in the include/compile path, wire it to the existing **`--permissive`** flag and make it a **warning, not an error** (and not a silent skip in the strict/default path). That keeps the default honest -- an unresolved label fails the build -- while giving an explicit, opt-in escape hatch.

Net: default = hard error; intentional optional refs = overlay; `--permissive` = warn-and-continue.

zeddii:
A clarification / correction to my `--permissive` suggestion above, now that we've pinned down what dtc actually allows.

`--permissive` **can't** make a concatenated `.dtsi` tolerate an unresolved label -- that's fatal at dtc *parse*, before any lopper leniency can run, and there's nothing to "skip" in a concatenated compile. The only way to get skip-with-warning is for the **assist to compile the fragment separately as a `/plugin/` overlay** (unresolved refs become `__fixups__`; `_unwrap_overlay_tree` skips the ones absent from the base). So `--permissive` only has meaning *if the assist does that conversion* -- it is not something the normal include/concat path can offer.

And that conversion has a hard cost, because the two dtc models are complementary, not superset/subset:

| Model | unresolved labels | delete base nodes/props |
|---|---|---|
| `.dtsi` (concatenated + recompiled) | ❌ fatal | ✅ `/delete-node/`, `/delete-property/` |
| `/plugin/` overlay (compiled standalone) | ✅ (fixups) | ❌ `/delete-node/` won't compile; `/delete-property/` no-ops on apply |

So the two capabilities are **mutually exclusive**:
- **Need to delete base nodes/properties** → it must stay a `.dtsi` (concatenated) → **no unresolved symbols**.
- **Need to tolerate unresolved/optional labels** → it must be an overlay → **no base deletions**.

No single fragment gets both. So the real per-fragment question is: does it need to *remove* base content, or *reference optional IP that may be absent*? It can't do both, and `--permissive` doesn't change that -- at most it would select the overlay path (gaining unresolved-tolerance, losing deletes).

kedareswararao:
@zeddii : The board dts contains all possible board-related information, such as the Ethernet node, flash info, eeprom info, and similar details. The purpose of passing this file to lopper assist is that, if a given design does not include Ethernet, the assist logic should remove the Ethernet node references from the user-provided board dts and, if needed, return an informational or warning message indicating that the relevant nodes were deleted because they are not mapped in the design. I hope this clarifies the use case; if not, please let me know and I’ll be happy to clarify further.

CC: @sivadur @dbingi-amd

zeddii:
Thanks @kedareswararao — to confirm the use case: the board `.dtsi` is a **superset reference** (every possible peripheral), and for a given design the entries whose IP isn't mapped shouldn't appear in the generated board DT, with a message noting what was left out.

Given that shape, the conclusion is clear: **this should be an overlay, and we should not pass it in alongside the system device tree for a concatenated merge** (the approach floated earlier in this thread). Here's why, and what "overlay" precisely means.

**"Overlay" = compiled standalone, not concatenated with the SDT.** The two paths have opposite failure behaviour:

- passed as a plain `-i` include (concatenated with the SDT and compiled as one unit) → a single `dtc` compile → an unresolved `&label` (optional IP absent from this design) is **fatal**: `Label or path ... not found` (even `-f` doesn't help).
- compiled **standalone** as `/plugin/` (`compile_overlay_standalone`) → `dtc` emits `__fixups__` instead of erroring → lopper unwraps and merges into the tree, **skipping** the fixups whose target is absent.

The skip-the-unmapped-IP behaviour you want exists only on the standalone-overlay path — which is exactly what @dbingi-amd moved to (`compile_overlay_standalone` → `_unwrap_overlay_tree` → `_merge_node_into_tree`). So: keep it there; do **not** route the superset file through `setup()`'s base concatenation, and do **not** reintroduce custom `.dtsi` parsing to do the removal.

**Skip vs. delete — the distinction that decides the mechanism:**

- **Leaving out an unmapped entry** (your ethernet case) = the overlay simply *not applying* that fragment (skip). No `/delete-node/` involved. This is what the use case needs, and the overlay path already does it.
- **Actually deleting a node that *is* in the base** (`/delete-node/`) is a different operation, and the two dtc models are mutually exclusive: an overlay **cannot** delete base nodes (won't compile), while a plain concatenated `.dtsi` **can** delete but can't tolerate unresolved labels. So if genuine base deletion is ever needed *alongside* optional-IP skipping, it has to be a **combination — the overlay for the optional content plus a plain `.dtsi` carrying the `/delete-*/` directives** — not custom parsing of one file.

**On the notice:** a summary of what was skipped (each dropped node + its unresolved label) is the right touch, so an *unexpected* omission (a typo, or an XSA missing IP it should have) is visible rather than silent.

Net:
- superset board file → **overlay, compiled standalone** (not passed with the SDT, not custom-parsed) → unmapped IP skipped, which is what's already implemented;
- genuine base deletion, if ever required → **overlay + a `.dtsi`**, never bespoke filtering.

kedareswararao:
> Thanks @kedareswararao — to confirm the use case: the board `.dtsi` is a **superset reference** (every possible peripheral), and for a given design the entries whose IP isn't mapped shouldn't appear in the generated board DT, with a message noting what was left out.
> 
> Given that shape, the conclusion is clear: **this should be an overlay, and we should not pass it in alongside the system device tree for a concatenated merge** (the approach floated earlier in this thread). Here's why, and what "overlay" precisely means.
> 
> **"Overlay" = compiled standalone, not concatenated with the SDT.** The two paths have opposite failure behaviour:
> 
> * passed as a plain `-i` include (concatenated with the SDT and compiled as one unit) → a single `dtc` compile → an unresolved `&label` (optional IP absent from this design) is **fatal**: `Label or path ... not found` (even `-f` doesn't help).
> * compiled **standalone** as `/plugin/` (`compile_overlay_standalone`) → `dtc` emits `__fixups__` instead of erroring → lopper unwraps and merges into the tree, **skipping** the fixups whose target is absent.
> 
> The skip-the-unmapped-IP behaviour you want exists only on the standalone-overlay path — which is exactly what @dbingi-amd moved to (`compile_overlay_standalone` → `_unwrap_overlay_tree` → `_merge_node_into_tree`). So: keep it there; do **not** route the superset file through `setup()`'s base concatenation, and do **not** reintroduce custom `.dtsi` parsing to do the removal.
> 
> **Skip vs. delete — the distinction that decides the mechanism:**
> 
> * **Leaving out an unmapped entry** (your ethernet case) = the overlay simply _not applying_ that fragment (skip). No `/delete-node/` involved. This is what the use case needs, and the overlay path already does it.
> * **Actually deleting a node that _is_ in the base** (`/delete-node/`) is a different operation, and the two dtc models are mutually exclusive: an overlay **cannot** delete base nodes (won't compile), while a plain concatenated `.dtsi` **can** delete but can't tolerate unresolved labels. So if genuine base deletion is ever needed _alongside_ optional-IP skipping, it has to be a **combination — the overlay for the optional content plus a plain `.dtsi` carrying the `/delete-*/` directives** — not custom parsing of one file.
> 
> **On the notice:** a summary of what was skipped (each dropped node + its unresolved label) is the right touch, so an _unexpected_ omission (a typo, or an XSA missing IP it should have) is visible rather than silent.
> 
> Net:
> 
> * superset board file → **overlay, compiled standalone** (not passed with the SDT, not custom-parsed) → unmapped IP skipped, which is what's already implemented;
> * genuine base deletion, if ever required → **overlay + a `.dtsi`**, never bespoke filtering.

@zeddii : Will test this pull request sometime this week and will come back with my test observations

kedareswararao (approved):
Have tested the changes it looks fine

kedareswararao:
@zeddii : Changes looks fine to me if you are okay with the changes please merge this, Dinesh has other pull requests which he needs to rebase on top of this

zeddii:
I don't have access to my test machine until Wednesday, I'll review and merge this then

PR #792, opened 2026-07-26
lopper: assists: Model TTC counters as child nodes

katrusatyasri:
Update the TTC node generation logic based on review feedback to model TTC counters as child nodes of the parent TTC device.

Changes
- Preserve TTC nodes as parent container nodes.
- Add #address-cells and #size-cells properties.
- Generate counter@0, counter@1, and counter@2 child nodes.
- Move interrupt properties from the parent node to the 
   corresponding child counter nodes.
- Update supported TTC properties for the new hierarchy.

This aligns the generated Devicetree hierarchy with the TTC hardware organization and the Zephyr TTC counter driver model.


katrusatyasri:
@kedareswararao ,
please review

kedareswararao (approved):
Changes looks fine to me

PR #793, opened 2026-07-27
lopper: assists: zephyr: Ensure nodes are proper for versal R5 zephyr-helloworld build

bentheredonethat:
SPI node needs right clock-frequency per inherited binding of https://docs.zephyrproject.org/latest/build/dts/api/bindings/flash_controller/cdns%2Cqspi-nor.html

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -1223,6 +1223,11 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine, options=None):
                             node["#size-cells"] = LopperProp("#size-cells")
                             node["#size-cells"].value = 0
                             node.add(node["#size-cells"])
+
+                        if node.propval("xlnx,clock-freq") != ['']:
+                            node + LopperProp(name="clock-frequency",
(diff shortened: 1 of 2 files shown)


bentheredonethat:
@kedareswararao  please review

this is needed for versal R5 Yocto SDT multiconfig target to build (both for configure step and compile step)

lopper/assists/gen_domain_dts.py
```diff
@@ -1223,6 +1223,10 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine, options=None):
                             node["#size-cells"] = LopperProp("#size-cells")
                             node["#size-cells"].value = 0
                             node.add(node["#size-cells"])
+
+                        node["clock-frequency"] = LopperProp("clock-frequency")
+                        node["clock-frequency"].value = node["xlnx,clock-freq"].value
```
kedareswararao: what if this property xlnx,clock-freq is not present in the node? please check for the existence of the property and default value in case if its absent

bentheredonethat:
@kedareswararao  addressed

kedareswararao (approved):

PR #795, opened 2026-08-03
Zephyr policy updates 20260727 

bentheredonethat:
Hi @zeddii  -- there are spec updates so user can specify Domain YAML linker and MPU characteristics for Zephyr

Importantly, gen_domain_dts does not generate MPU permissions or linker scripts. It preserves the results produced by the dedicated assists.

Hi @sivadur  @kedareswararao  there are the gen-domain changes for:

  - /scratch/blevinsk/lopper/lopper/assists/gen_domain_dts.py                                                                                                                                         
  - /scratch/blevinsk/lopper/lopper/assists/zephyr_supported_comp.yaml                                                                                                                                
                                                                                                                                                                                                      
  It does the following:                                                                                                                                                                              
                                                                                                                                                                                                      
  - Preserves nodes explicitly marked with:                                                                                                                                                           
                                                                                                                                                                                                      
    compatible = "zephyr,memory-region";                                                                                                                                                              
                                                                                                                                                                                                      
  - Preserves nodes referenced from /chosen, including:                                                                                                                                               
                                                                                                                                                                                                      
    zephyr,sram = "...";                                                                                                                                                                              
    zephyr,ipc_shm = "...";                                                                                                                                                                           
                                                                                                                                                                                                      
  - Prevents the generic Zephyr domain transformation from overwriting an already selected zephyr,sram.                                                                                               
  - Renames R5/R52 memory nodes to match their processor-local reg addresses and updates /chosen references after the rename.                                                                         
  - Preserves these generated properties during pruning:                                                                                                                                              
                                                                                                                                                                                                      
    zephyr,memory-region                                                                                                                                                                              
    zephyr,memory-attr                                                                                                                                                                                
    zephyr,sram                                                                                                                                                                                       
                                                                                                                                                                                                      
  - Normalizes ZynqMP IPI mailbox properties for Zephyr:                                                                                                                                              
                                                                                                                                                                                                      
    xlnx,ipi-id → local-ipi-id                                                                                                                                                                        
    xlnx,ipi-id → remote-ipi-id                                                                                                                                                                       
    reg-names = "host_ipi_reg"                                                                                                                                                                        
    status = "okay"                                                                                                                                                                                   
                                                                                                                                                                                                      
  - Removes incompatible properties from destination mailbox children where required.                                                                                                                 
  - Extends the supported-compatible policy for:                                                                                                                                                      
      - zephyr,memory-region                                                                                                                                                                          
      - mmio-sram                                                                                                                                                                                     
      - zephyr,mbox-ipm                                                                                                                                                                               
      - vnd,mbox-consumer                                                                                                                                                                             
      - xlnx,zynqmp-ipi-mailbox                                                                                                                                                                       
      - xlnx,zynqmp-ipi-dest-mailbox    
      
      
      
  ## Intended execution order

  Domain YAML
      ↓
  YAML expansion
      ↓
  zephyr-mpu
      ├── emits zephyr,memory-region
      ├── emits zephyr,memory-attr
      └── selects /chosen/zephyr,sram
      ↓
  zephyr-linker
      ├── writes the linker script
      └── writes the layout report
      ↓
  gen_domain_dts
      ├── prunes the domain
      ├── preserves generated Zephyr metadata
      ├── fixes local RPU memory names
      └── produces the final Zephyr DTS


zeddii:
I'm out of the office, back on Wednesday and will review these then.

bentheredonethat:
@zeddii  ok thank you!

zeddii:
Thanks @bentheredonethat — this is well-structured and the split is right (linker/MPU in dedicated assists, `gen_domain_dts` only preserving the generated metadata). CI is green and the new tests target the tricky bits (sram-override guard, ipc_shm refs, RPU local-TCM range, linker expand +/-). A few items before it can be merged:

**1. Rebase needed.** #790 just merged, and it touches `gen_domain_dts.py` in the same areas this PR does — please rebase on latest `master`.

**2. Rename may leave stale references to the renamed node.** It looks to me like `xlnx_zephyr_fixup_rpu_memory_names()` only fixes `/chosen` refs for the hardcoded `("zephyr,sram",)`. A reference to a renamed node could be left pointing at the old path via:
- other `/chosen` entries (e.g. `zephyr,ipc_shm`) — `_zephyr_node_must_be_preserved()` scans *all* `/chosen` props;
- a `&label`/phandle reference from another node — the rename does `delete` → mutate `abs_path`/`name` → `add`, and `add`/`delete` don't re-register `__symbols__`/phandles on their own.

  To refresh after the rename: call `sdt.tree.resolve()` (re-resolves phandles/refs — the same call `assemble_sdt.py` uses after structural edits), and update the renamed node's `/__symbols__` entry to the new path — there's already a precedent for exactly this in this file (the domain `/__symbols__` fix-up, ~L655–668), followed by `sdt.tree.sync()`. A test that renames a node referenced by both `zephyr,ipc_shm` and a `&label` elsewhere, asserting both still resolve, would confirm it.

**3. Guard the IPI id access.** In `xlnx_remove_unsupported_nodes()`, `node["xlnx,ipi-id"].value` is read after matching `xlnx,zynqmp-ipi-mailbox` / `-dest-mailbox` — if a matching node lacks `xlnx,ipi-id` this throws `KeyError`. A cheap `if node.props("xlnx,ipi-id")` guard would harden it.

**Minor (non-blocking):** the rename uses `reg[1]` as the new unit-address, which assumes the high address cell is 0 (local TCM view) — fine for the intended case, worth a one-line comment.

With the rebase + items 2 and 3 addressed, this is good to go.

bentheredonethat:
@zeddii  rebased , updated , pushed!

PR #796, opened 2026-08-04
petalinuxconfig_xlnx.py: emit disabled UARTs under 'disabled_slaves'

Rajukumar45:
QEMU instantiates a serial port for every UART on the SoC, including disabled ones, and numbers them serial0, serial1, ... by base address. For gen-machine-conf to build a matching QEMU '-serial' line, the syshw data must also list the disabled UARTs.

Detect disabled UART nodes the same way as enabled devices (by matching their 'compatible' against the schema's serial driver list) and route them through the existing serial-slave handling into a dedicated 'disabled_slaves' section. Keeping them out of 'slaves' ensures nothing that iterates the slaves (e.g. the Kconfig serial menu) is affected.

- Accumulate the serial-compatible set during the existing device_type_map build pass instead of a second loop.
- Omit 'ip_name' when a node has no 'xlnx,ip-name' (disabled UARTs are not annotated by the Xilinx SDT flow) rather than emit an empty value.

### DIFF
--- a/lopper/assists/petalinuxconfig_xlnx.py
+++ b/lopper/assists/petalinuxconfig_xlnx.py
@@ -312,6 +312,9 @@ def xlnx_generate_petalinux_config(tgt_node, sdt, options):
 
         # Build device type mapping for non-processor/non-memory devices
         device_type_map = {}
+        # Serial driver compatibles, accumulated in the same pass; used below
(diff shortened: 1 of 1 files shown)


Rajukumar45:
No extra reviews are necessary.

PR #797, opened 2026-08-06
overlays: separate conditional property activation

bentheredonethat:
Add a standalone conditional_properties assist that selects a merged overlay tree without applying domain access filtering or pruning. Support explicit condition selection and activation through a target node's lopper,activate property.

Share the activation helper with domain_access to preserve its existing behavior, document both workflows, and cover standalone activation across the embedded-overlay two-pass flow.

### DIFF
--- a/docs/conditional-properties.md
+++ b/docs/conditional-properties.md
@@ -108,6 +108,18 @@ ref-counting, memory filtering, chosen node injection, and pruning.  If
 `lopper,activate` is absent, `os,type` is used as a fallback so existing
 domain YAML files work without modification.
 
+To retain devices outside the selected domain while still activating its
+conditional properties, disable only the device-pruning phase:
+
+```bash
+lopper system-top.dts output.dts -- \
+    domain_access -t /domains/linux-domain --no-device-prune
+```
+
+This option skips refcount-based CPU, bus, and device removal. Domain memory
+processing and chosen-node handling still run. Without the option,
+`domain_access` retains its normal domain-scoped output behavior.
+
 ---
 
 ## Per-Domain Driver Binding ("OpenAMP Pattern")
--- a/lopper/assists/domain_access.py
+++ b/lopper/assists/domain_access.py
@@ -179,8 +179,10 @@ def usage():
     print( """
    Usage: domain_access [OPTION]
 
-      -p       permissive matching on target node (regex)
-      -v       enable verbose debug/processing
+      -p, --permissive       permissive matching on target node (regex)
+      -t, --target NODE      apply the specified domain
+      --no-device-prune      retain devices outside the selected domain
+      -v, --verbose          enable verbose debug/processing
 
     """)
 
@@ -223,9 +225,14 @@ def core_domain_access( tgt_node, sdt, options ):
         args = []
 
     # --permissive means that non-SMID devices/memory will be consulted
-    opts,args2 = getopt.getopt( args, "vt:p", [ "verbose", "target=", "permissive" ] )
+    opts,args2 = getopt.getopt(
+        args,
+        "vt:p",
+        ["verbose", "target=", "permissive", "no-device-prune"],
+    )
 
     permissive = False
+    device_prune = True
     command_line_target=""
     for o,a in opts:
         if o in ('-v', "--verbose"):
@@ -234,6 +241,8 @@ def core_domain_access( tgt_node, sdt, options ):
             permissive = True
         elif o in ("-t", "--target"):
             command_line_target = a
+        elif o == "--no-device-prune":
+            device_prune = False
         elif o in ('-h', "--help"):
             # usage()
             sys.exit(1)
@@ -509,7 +518,7 @@ def core_domain_access( tgt_node, sdt, options ):
 
     if cpu_prop:
         refd_cpus, unrefd_cpus = lopper_lib.cpu_refs( sdt.tree, cpu_prop, verbose )
-        if refd_cpus:
+        if refd_cpus and device_prune:
             ref_nodes = sdt.tree.refd( "/cpus.*/cpu.*" )
 
             # now we do two types of refcount delete
@@ -594,9 +603,11 @@ def core_domain_access( tgt_node, sdt, options ):
                return False
            """
 
-    _info( f"core_domain_access: filtering on:\n------{code}\n-------\n" )
-
-    sdt.tree.filter( "/", LopperAction.DELETE, code, None, verbose )
+    if device_prune:
+        _info( f"core_domain_access: filtering on:\n------{code}\n-------\n" )
+        sdt.tree.filter( "/", LopperAction.DELETE, code, None, verbose )
+    else:
+        _info("core_domain_access: device pruning disabled")
 
     # filter #2:
     #    - starting at simple-bus nodes
@@ -606,7 +617,7 @@ def core_domain_access( tgt_node, sdt, options ):
     #    - starting at reserved memory parent
     #    - drop any unreferenced elements
 
-    for n in nodes_to_filter:
+    for n in nodes_to_filter if device_prune else []:
         code = """
                p = node.ref
                if p <= 0:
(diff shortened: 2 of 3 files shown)


bentheredonethat:
@zeddii  this is to enable the (very useful!) feature of conditional properties WITHOUT requiring the domain access plugin pruning.

Note original functionality is still preserved.

zeddii:
The implementation is wrong, and it conflicts with some other changes that are happening to allow sigils to be stacked.

This doesn't belong in an assist at all, but is core functionality.

Once I have the new conditionals working more, I Can reply in detail. But as-is, this won't merge.

zeddii:
Thanks @bentheredonethat. I want to step back on the approach before this goes further.

As you know, conditional/sigil activation is **already part of applying a domain**: `core_domain_access()` activates the named overlay (`lopper,activate`, falling back to `os,type`) as its first step, then refcounts what the domain declares, prunes the rest, applies memory, and injects `chosen`. That pruned, OS-specific device tree *is* the deliverable — an OS-specific tree with the other parts removed is the whole point of running lopper on a system SDT. So for the `timer@f1e90000` / `compatible!linux: uio` case, running `domain_access` on `APU_Linux` already gives both the Linux binding and the tree scoped to that domain.

And `domain_access` doesn't prune anything gratuitously — it acts only on what the input domain declares. So: what does the standalone (uncropped) activation give you that the domain-applied OS-specific tree doesn't? If some phase (the prune, the memory filtering, the `chosen` injection) is doing something you don't want for this output, let's pin down *which* — that's the useful signal.

To be clear, I won't take this as a standalone `conditional_properties` assist. Every assist that has to be explicitly invoked is another failure point in the pipeline — one more step to forget, to order wrong, or to let drift out of sync. Beyond that, two specifics:

1. **It would collide with the sigil/overlay-handling work landing around September.** Rather than add a parallel activation path now and rework it then, let's solve your need as part of that.
2. **There's a cleaner fit:** make the phases of `domain_access` individually selectable — the same "activate this / skip that" spirit we already use for sigils. If a particular step (prune / memory / chosen) is what's in your way, a caller runs `domain_access` and turns *that* phase off, instead of reimplementing activation elsewhere. Activation stays where it belongs (part of applying a domain), it's one code path, and you get explicit "activate without X" behavior.

So the plan: we'll hold this PR. If there's a genuine near-term gap, we find the smallest change to cover it (the phase-toggle on `domain_access` is the likely shape) — then address it more fully with the new conditionals and APIs in the September timeframe. If you can describe the exact output you need, and which phase is the blocker, that'll tell us whether a small change is even needed now.

bentheredonethat:
@zeddii  thanks for **very** constructive feedback and comment on usage

I will amend this PR as you have noted - where domain access can operate without the pruning nodes via access list portion. 

thank you!!

bentheredonethat:
@zeddii 
The phase blocking my output is device-tree pruning: I need the condition-resolved Linux bindings while retaining the complete system tree. I’ll revise this to remain entirely within domain_access, adding --no-prune to skip only its refcount/device-pruning phase. Conditional activation will continue through the existing lopper,activate/os,type path, while memory and chosen processing remain unchanged. Default behavior will be unaffected.

zeddii:
Thanks @bentheredonethat — moving this into `domain_access` as a phase toggle is the right direction, and the command-line option is a legitimate way to drive it. The blocker is that it can't *only* be a command-line option.

As you know, `domain_access` doesn't only run when it's explicitly invoked with `-- domain_access …` — it also auto-fires on the `access-domain,domain-v1` compatible, and it runs inside pipelines. A `--no-device-prune` flag is invisible in both of those paths, so whatever controls this has to work as **data in the tree**, not only on the command line. The flag can stay as an override; it just can't be the only way.

The clean, data-driven trigger for "activate but keep everything" is the spec's own glob. `access: "*"` already means "everything" (chapter 7). **We're writing the changes to extend that across the whole need:**
- adding globbing to `memory` (a small, symmetric spec update) so `memory: "*"` means all physical memory, and
- having `domain_access` treat a `"*"` glob with no parent pool as "keep all" instead of the current error.

With those, a domain carrying `access: "*"` and `memory: "*"` (plus its `os,type`) yields the full device tree with the overlays activated — and because it's data in the domain, it behaves identically whether auto-fired, run in a pipeline, or invoked explicitly. Those commits and the spec update are in progress on our side.

Two things first:

1. **What's the actual goal?** Why do you need the conditionals activated *without* the domain processing (the device/memory scoping)? What consumes that output? That reason tells us whether this is even the right approach — and whether the memory (and chosen) phases matter for your case too.
2. **Regardless of the reason,** the best way to express "give me the whole device tree + overlays" is a **separate domain** carrying the `"*"` globs, and you trigger *that* domain when you want the full view — rather than telling a scoped domain to skip the scoping it declares (self-contradictory), or adding a standalone assist. One domain describes "the Linux-scoped view," another describes "the whole system in the Linux flavor"; both are just domains, applied the same way.

So let's hold this PR while we land the `memory` glob and the no-parent-`"*"` keep-all change, and nail down the goal — then your case becomes a domain with globs, driven by data, working in every path.

bentheredonethat:
@zeddii  the lopper conditional activate is so i can pass zephyr specific or linux specific flags

the whole access / * thing is fine... but it does require a parent domain to describe all nodes and then a child domain  to use the *  char in its access if I understand that one correctly

regardless - I just want to be able to have some nodes have property X or Y for zephyr domain and Z for linux domain

I have solved it for my openamp specific use cae with the PR https://github.com/devicetree-org/lopper/pull/798 so if you are OK with that self-contained one it is suffcient for the libmetal and openamp use cases at least and your generic solution can then be done as needed in your more generic solutiion

is that ok ?

zeddii:
Thanks @bentheredonethat. The keep-all changes are on master now — please pull and test your case against this model.

**On the parent/child point:** `access: "*"` no longer needs a parent domain describing all the nodes. As you know, that used to be the only way `"*"` resolved — matched against a parent pool. We changed `domain_access` so an unscoped `"*"` (no parent) means "keep everything," so a single standalone domain does what you want. Same for `memory: "*"` (all physical memory) and `sram: "*"`.

**Your case — condition-resolved Linux bindings with the complete system tree retained — is one domain:**

```yaml
axi:
  timer@f1e90000:
    compatible: "cdns,ttc"        # base: non-linux domains
    compatible!linux: "uio"       # linux overlay replaces it

domains:
  APU_Linux:
    compatible: openamp,domain-v1
    os,type: linux                # activates the 'linux' sigils
    access:
      - dev: "*"                  # keep all devices (no pruning)
    memory:
      - "*"                       # keep all physical memory
    sram:
      - "*"
```

One command:

```bash
lopper.py -f --permissive --auto \
    -i domains.yaml system-top.dts out.dts \
    -- domain_access -t /domains/APU_Linux
```

On our selftest SDT the conditional timer's `compatible` goes `cdns,ttc` → `uio` (linux sigils applied) and the full tree is retained — all device nodes and all physical memory, nothing pruned. That's exactly "conditionals resolved, complete system tree kept," through the existing activation path — no `--no-prune` flag and no separate assist.

**And `"*"` is just the "keep everything" end of the range — you're not tied to it.** Each domain is independent data, so you can define several domains over the same system and run whichever you want: one with `"*"`, one with an explicit access list, one that lists most devices but deliberately leaves a few out. If a device is perfectly valid but you just don't want it in a given run, scope that domain's access to omit it and target it with `-t`; the other domains sit in the same file untouched. `"*"` for the views that want everything, explicit lists for the views that don't — same mechanism, same pass.

Because it's data on the domain it also composes: it works auto-fired, explicitly, or mid-pipeline, and the keep-all intent is recorded as a persisted `lopper,*-keep-all` property so a later stage that re-consumes the tree stays idempotent. It isn't OS-interpreted; strip it with a delete lop at finalize if you want a clean deliverable.

**On #798:** that's a different, contained change — openamp timer trimming, keeping UIO in the linux case — and it isn't the parallel-activation concern I had here. I'll look at it on its own.

If the model above covers your case once you've tested it, we can close this in favor of it, and the broader sigil/keep-all stacking work in the September timeframe builds on exactly this.

zeddii:
I inadvertently re-implemented the command line options, they were supposed to be proof of concept. So I'll double check the test coverage, but otherwise, we should be good on this. Once you confirm.

bentheredonethat:
@zeddii  confirmed this works for me - thanks

bentheredonethat:
closing

PR #798, opened 2026-08-07
openamp: xlnx: clean up timer trimming

bentheredonethat:
in linux case - keep UIO if present. thats it.
else (non linux cases) just make sure all extent timers have the correct compatible

### DIFF
--- a/lopper/assists/openamp_xlnx.py
+++ b/lopper/assists/openamp_xlnx.py
@@ -127,28 +127,29 @@ def xlnx_openamp_trim_timers(sdt, target_os, machine):
     timer_pvals = [ n.propval("timer") for n in domains[0].subnodes(children_only=True) if n.propval("timer") != [''] ]
 
     # only do trim if timer prop is provided
-    if not timer_pvals:
+    if timer_pvals == []:
         return False
-
-    flattend_timer_pvals = [item for sublist in timer_pvals for item in sublist]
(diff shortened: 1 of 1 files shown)


zeddii:
Thanks @bentheredonethat — this is a reasonable contained cleanup for the openamp timer case, and the NOTE is exactly right: once the sigil / conditional-property work is in, this routine folds into that. CI is green and it's mergeable. One thing to fix and one small ordering nit before it goes in.

**1. The removed `if not domains: return False` guard.** The diff drops that guard but still does `domains[0].subnodes(...)` immediately after — so if `domains` comes back empty (no domain matches the cpu), this now raises `IndexError` instead of returning cleanly. Was the removal intentional? If `domains` can still be empty on any path, please keep an early return before indexing `[0]`.

**2. linux branch — delete-then-enable ordering.**
```python
for node in tree["/axi"].subnodes(children_only=True, name="timer@*"):
    if node not in relevant_timer_nodes:
        tree.delete(node)
    if "uio" in node.propval("compatible"):
        node["status"] = "okay"
```
The `uio`/status check runs even for a node just deleted on the line above (it reads/sets properties on a removed node). Making the enable an `elif`, or enabling before deleting, avoids operating on a deleted node. Also note this now considers *all* `timer@*` nodes for deletion, where the old code only deleted `cdns,ttc` timers — so an unreferenced UIO timer would now be removed too; if that's intended, fine, just calling it out.

Otherwise the approach is good, and I'm happy to take it as the openamp-specific holdover until the generic sigil handling lands.

PR #799, opened 2026-08-11
lopper: assists: gen_domain_dts: auto-detect carrier-board peripheral…

kedareswararao:
… Kconfig

Lopper currently generates the Zephyr-specific devicetree but not the Kconfig defaults a peripheral may also need, so those are hand-added per board today. Add infra to generate these too: if a peripheral needs certain configs, add a rule to emit them into the board's Kconfig.defconfig, so it gets copied to the board and consumed just like the generated DT. Board-specific entries (e.g. EEPROM) can be added by reading the board name off the SDT root node.

Also guard the board_Kconfig.defconfig scratch file, reused across west lopper-command invocations, against being appended to under a stale board's guard left over from a prior run.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -835,7 +835,13 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
             delete_unused_props( sdt.tree[match_cpunode] , driver_proplist, False)
 
     if zephyr_dt:
-        if "r52" in machine or "a78" in machine or "a72" in machine or "r5" in machine or "a53" in machine:
+        is_arm = "r52" in machine or "a78" in machine or "a72" in machine or "r5" in machine or "a53" in machine
+        board_symbol = board_symbol_for_machine(machine) if is_arm else "BOARD_MBV32"
+
+        if board_symbol:
+            _reset_board_kconfig_scratch(sdt.outdir)
+
+        if is_arm:
             xlnx_generate_zephyr_domain_dts_arm(tgt_node, sdt, options, machine)
             _xlnx_zephyr_assign_ttc0(sdt.tree, machine)
             if "a78" in machine or "a72" in machine:
@@ -883,6 +889,12 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
                 print(f"[ERROR] Failed to process overlay file: {e}")
                 import traceback
                 traceback.print_exc()
+
+        # Peripheral Kconfig detection must run after the board overlay merge
+        # above -- that's what actually adds nodes like the I2C EEPROM/mux or
+        # UFS controller to sdt.tree for carrier-board-specific designs.
+        if board_symbol:
+            _append_dt_peripheral_kconfig(sdt, board_symbol)
 
     return True
 
@@ -1058,6 +1070,19 @@ def xlnx_generate_zephyr_domain_dts_arm(tgt_node, sdt, options, machine):
         if node.name == 'reserved-memory' and 'r52' in machine:
             node.delete('ranges')
             node + LopperProp(name='ranges')
+
+    board_symbol = board_symbol_for_machine(machine)
+    if board_symbol:
+        default_entries = _ARM_BOARD_DEFAULT_KCONFIG.get(board_symbol, [])
+        if default_entries:
+            content = _KCONFIG_LICENSE_HEADER
+            content += f"if {board_symbol}\n\n"
+            content += _render_kconfig_entries(default_entries)
+            content += f"endif # {board_symbol}\n"
+
+            board_defconfig_file = os.path.join(sdt.outdir, "board_Kconfig.defconfig")
+            with open(board_defconfig_file, 'w') as board_defconfig:
+                board_defconfig.write(content)
 
     return True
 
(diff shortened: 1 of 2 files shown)


kedareswararao:
@zeddii : Please review when you have sometime

zeddii:
Thanks @kedareswararao — auto-generating the DT-derived Kconfig defaults instead of hand-adding them per board is a useful direction. A few things to sort before it can be merged: one correctness bug, a test gap, and something I'd like to understand.

**1. Blocking — the stale-file guard has a hole (same board, silent config bleed).**

`board_Kconfig.defconfig` lives in `sdt.outdir`, which is reused across `lopper_command.py` runs. Both write paths are conditional, and neither touches the file when a run has nothing to add:
- Path A (`gen_domain_dts.py:1069-1079`) writes a fresh file only `if default_entries:` — a board not in `_ARM_BOARD_DEFAULT_KCONFIG` gets `[]` and skips the write.
- Path B (`_append_dt_peripheral_kconfig:1794`) does `if not entries: return` *before* the guard in `_append_kconfig_entries_before_endif` can run.

So the guard only protects the append case. For a board not in the dict (e.g. `versal2_apu`):

```
Run 1: versal2_apu carrier design WITH an i2c EEPROM  -> Path B creates
         `if BOARD_VERSAL2_APU ... <eeprom cfg> ... endif`
Run 2: versal2_apu plain design, NO EEPROM (same outdir)
         Path A: [] -> no write;  Path B: no entries -> early return
         -> file still holds Run 1's block; guard is TRUE (same board)
         -> the plain design silently inherits an EEPROM Kconfig default it never declared.
```

Fix: reset the file for the current board unconditionally at the start of the run (or delete any pre-existing scratch copy) before either path, so a prior run can't survive regardless of whether this run adds anything.

**2. Please add tests.** The new helpers (`detect_dt_peripherals`, `board_symbol_for_machine`, `detect_board_i2c_mux_prio`, `_append_kconfig_entries_before_endif`) are pure string/regex logic and easy to unit-test — especially the stale-file guard above, which is the crux of the change and currently unverified.

**3. Not a blocker — just thinking out loud on `_ARM_BOARD_DEFAULT_KCONFIG`.** I was wondering about the baked-in defaults here — they look like a verbatim copy of upstream's `boards/amd/versal_apu/Kconfig.defconfig` and `versal_rpu/Kconfig.defconfig` (the `BUILD_OUTPUT_BIN` default, the `USERSPACE`/`COMPILER_ISA_THUMB2` block), which upstream also maintains, so I could see them drifting over time. I'm not exactly sure how these files get consumed by the Zephyr build on the `lopper_command.py` side (copied over the in-tree file, merged, sourced?), so this really isn't blocking — I just want to understand the flow, and maybe see if there's a smaller-footprint way to get the same result (e.g. emitting only the DT-derived additions and leaving the upstream baseline in place). If you can sketch how the generated `board_Kconfig.defconfig` gets picked up, that'd help.

None of this is a rejection — the peripheral-detection logic is sound and well-commented, and the blast radius is safely Zephyr-scoped. The stale-file bug + tests are the things to fix in this PR; #3 is just me trying to understand the flow.

kedareswararao:
@zeddii : Thanks for the review  Pushed a fix for #1 and #2 

  #1 — fixed. You're right, the guard only protected the append path. Added _reset_board_kconfig_scratch(), called
  unconditionally at the top of the zephyr_dt dispatch before either write path runs, so a prior run's content for
  board_Kconfig.defconfig can never survive into a run that has nothing new to contribute — regardless of which path
  would otherwise have (not) touched the file.

  #2 — tests added. tests/test_zephyr_peripheral_kconfig.py covers board_symbol_for_machine, detect_board_i2c_mux_prio,
  detect_dt_peripherals (including the disabled-node and dedup cases), _append_kconfig_entries_before_endif (including
  the stale-different-board-guard case), and _reset_board_kconfig_scratch, with a regression test reproducing the
  exact cross-run bleed scenario you described. Also verified the fix end-to-end through the real west lopper-command
  pipeline: pre-seeded the scratch workspace with stale if BOARD_MBV32 content and ran it against a board with no
  baked-in defaults (versal2_rpu, via vek385_revb_rpu_ced) — confirmed the stale content is gone and only the
  correctly-guarded, genuinely-detected entries survive.

  #3 — the flow, and the smaller-footprint question. On the Zephyr side, scripts/west_commands/lopper_command.py's ARM
  handlers do a plain blind copy: if workspace/board_Kconfig.defconfig exists after the lopper run, it's copied
  straight over boards/amd/<board>/Kconfig.defconfig — no merge, no diff. That's why _ARM_BOARD_DEFAULT_KCONFIG bakes
  in the existing hand-authored content (BUILD_OUTPUT_BIN, the USERSPACE/COMPILER_ISA_THUMB2 block) verbatim: a blind
  copy of just the DT-derived additions would otherwise wipe those out.

  We considered the merge-based alternative you're describing (read the existing in-tree file, splice in just the new
  entries, on the lopper_command.py side) and moved away from it deliberately — it pushes Kconfig-parsing/merge logic,
  and its own edge cases (nested if blocks, idempotency across repeated runs), into the West command, for not much
  benefit given only 4 boards currently have any baked-in content at all (versal_apu, versal_rpu, zynqmp_apu,
  zynqmp_rpu), and it's small and stable. You're right that it can drift if someone hand-edits one of those upstream
  files without updating the table to match — that's a real, accepted tradeoff for simplicity, not a gap we missed.
  Happy to reconsider if you'd rather see the merge-based approach.

 the current lopper_command.py does shutil.rmtree(workspace) at the end of a successful run, which is a second, independent line of defense against this same class of scratch-file staleness on the Zephyr side. The Lopper-side reset is still worth keeping regardless — it covers a run that fails  before reaching that cleanup, and any other consumer of gen_domain_dts.py that doesn't clean up the same way.
 
 The same files generated being used in Yocto flows gen-machine-conf tool as well

zeddii:
Fair enough. The rmtree at the end of the run covers the common flow, and the hardcoded table is a reasonable tradeoff to own on your side. No objection from me — good to merge.

PR #800, opened 2026-08-11
lopper: assists: fix Zephyr combined-DTS overlay phandle merge

dbingi-amd:
PR #790 merges the board .dtsi from the SDT folder directly into the domain tree inside Lopper, instead of emitting a separate board.overlay for Zephyr to apply later. That combined-DTS approach is correct, but it exposed a gap in what the existing lopper overlay APIs actually resolve.

When dtc compiles a plugin overlay, it records two kinds of phandle fixups. Global fixups under __fixups__ (often 0xffffffff pointing at a fragment target like &gem0) are handled by _unwrap_overlay_tree() and _resolve_overlay_fixups(). Local fixups
under __local_fixups__ (references to labels defined inside the same fragment, like phy-handle = <&phy0>) are not handled by those APIs at all.

That did not matter in the old flow, because Zephyr applied the overlay at build time and dtc fixed both kinds in one pass. After #790, the merge happens inside Lopper on an in-memory domain tree, so local fixups never get applied unless we do it explicitly. On top of that, _merge_node_into_tree() resolves properties immediately after copy, which can bind wrong phandle values before the tree is ready, and _unwrap_overlay_tree() leaves nested overlay children on /fragment@N/__overlay__/... paths instead of real domain paths.

This commit adds three helpers in zephyr_board_dt.py and reorders _merge_plugin_overlay() so local phandles are fixed after merge, without changing Lopper core.
1. _remap_unwrapped_overlay_nodes() rewrites overlay node paths from fragment space (/fragment@0/__overlay__/...) to real domain paths before merge, so nested nodes land in the right place.
2. _merge_overlay_node_deferred() replaces _merge_node_into_tree() for this path. It copies overlay properties onto domain nodes but skips immediate prop.resolve(), so phandles are not finalized too early.
3. _finalize_plugin_overlay_phandles() applies what core does not: it uses overlay __symbols__ to register labels on domain nodes, reads __local_fixups__ to find which properties need rewriting, runs main_tree.resolve(), then replaces temporary
overlay phandle numbers with the correct domain phandles — only for properties dtc flagged in __local_fixups__, so unrelated integer properties are left alone.

### DIFF
--- a/lopper/assists/zephyr_board_dt.py
+++ b/lopper/assists/zephyr_board_dt.py
@@ -113,6 +113,26 @@ def _merge_root_plugin_nodes(overlay_tree, main_tree, fragment_overlay_to_real):
                 pass
 
 
+def _apply_overlay_symbol_labels(main_tree, overlay_tree, fragment_map):
+    """Register overlay /__symbols__ labels on merged domain-tree nodes."""
+    try:
+        symbols = overlay_tree["/__symbols__"]
+    except (KeyError, Exception):
+        return
+
+    for label, prop in symbols.__props__.items():
+        sym_path = prop.value[0] if isinstance(prop.value, list) else prop.value
+        real_path = _rewrite_fragment_path(sym_path, fragment_map)
+        if not real_path:
+            continue
+        try:
+            node = main_tree[real_path]
+        except (KeyError, Exception):
+            continue
+        node.label = label
+        main_tree.__lnodes__[label] = node
+
+
 def _merge_plugin_overlay(content, main_tree, sdt, sdt_folder, work_dir):
     """Merge &label { } board fragments via plugin compile + unwrap (optional HW safe)."""
     dtso_path = os.path.join(work_dir, "board_zephyr.dtso")
@@ -137,6 +157,7 @@ def _merge_plugin_overlay(content, main_tree, sdt, sdt_folder, work_dir):
         _resolve_overlay_fixups(main_tree, fixups, local_fixups)
 
     fragment_map = _fragment_overlay_to_real(overlay_tree, main_tree)
+    _apply_overlay_symbol_labels(main_tree, overlay_tree, fragment_map)
     _merge_root_plugin_nodes(overlay_tree, main_tree, fragment_map)
 
     labels = sorted({n.label for n in nodes if getattr(n, "label", None)})


dbingi-amd:
@kedareswararao, please review the changes

kedareswararao (approved):

zeddii:
Thanks @dbingi-amd — your diagnosis was right: after #790 merges the board `.dtsi` in-memory, dtc's `__local_fixups__` (in-overlay `&label` references) were never re-based, because core only handled the global `__fixups__`. That's a real gap, and it turned out to be a core one rather than something the Zephyr assist should carry — so I've added it to core (on master):

- `_unwrap_overlay_tree` now parses `__local_fixups__` into `(holder, prop, byte_offset, target)` entries, and `_resolve_overlay_local_fixups` rewrites the referenced cell **by dtc byte offset** (`idx = byte_offset // 4`) to the target overlay node's phandle in the merged tree. Because it's offset-based, a specifier cell in a multi-cell property (`clocks = <&clk0 5>`, `gpios`, `interrupt-map`) is never touched — the value-match approach in `_finalize_plugin_overlay_phandles` would silently re-bind those.
- `_resolve_overlay_fixups` now resolves global + local in one call, so a caller that unwrapped an overlay can't drop the local set.
- It runs implicitly wherever an overlay is applied (like the global fixups already do), including the Zephyr combined-DTS path — `zephyr_board_dt.py` on master already passes `local_fixups` through, so once you rebase, the in-overlay references resolve for free.

So on this PR you can drop the parallel path entirely — `_finalize_plugin_overlay_phandles`, `_merge_overlay_node_deferred`, and `_remap_unwrapped_overlay_nodes` — and rely on the core resolution (same play as #803's `rename()`). If there's a case the core path doesn't cover once you've rebased, point me at it and I'll extend it.

Thanks again — these keep surfacing gaps worth fixing at the root.

kedareswararao (approved):

dbingi-amd:
@zeddii Thanks for the core __local_fixups__ support — that covers in-fragment phandle fixups in the combined-DTS merge path.

One piece was still missing for combined-DTS parity with build-time overlay merge: node labels from /__symbols__. Your change fixes phandle cells; dtc still records overlay labels separately under /__symbols__. 
When we merge board fragments in gen_domain_dts, those labels were not applied on the merged tree, so output could have the right nodes and properties but lose labels such as storage_partition:.

Fix: _apply_overlay_symbol_labels() in zephyr_board_dt.py, called from _merge_plugin_overlay() after fixup resolution and before _merge_root_plugin_nodes(). It maps each /__symbols__ entry through the fragment overlay prefix to the real domain path and registers the label on the merged node. Label registration only — no extra phandle logic.

zeddii:
Thanks — good catch on the labels. I've pulled that into core too (on master): the overlay merge now applies the `/__symbols__` labels alongside the phandle fixups, so it's handled for every consumer, not just the combined-DTS path.

Can you do a quick rebase and drop `_apply_overlay_symbol_labels`? It should fall out the same way the phandle helpers did.

dbingi-amd:
Closing — superseded by core overlay fixes 98a612f (https://github.com/devicetree-org/lopper/commit/98a612f) (__local_fixups__ resolution by offset) and 2bdb783 (https://github.com/devicetree-org/lopper/commit/2bdb783) (/__symbols__ label application on merge).

These resolve the incorrect in-fragment phandles (phy-handle) and missing overlay labels (storage_partition:). The assist-side work in this PR is no longer required.

Thanks @zeddii and @kedareswararao for the review.

PR #802, opened 2026-08-11
gen_domain: preserve Zephyr IPI references during renaming

bentheredonethat:
Rename ZynqMP IPI child nodes from their register addresses so the generated unit addresses match reg. Apply the corresponding child@reg naming to Versal IPI destination mailboxes during Zephyr conversion.

Preserve symbolic, alias, chosen, and other path references when renaming ZynqMP or Versal nodes. Numeric phandle references remain unchanged.


bentheredonethat:
@kedareswararao  please review

bentheredonethat:
.

PR #803, opened 2026-08-12
gen_domain: preserve Zephyr IPI references during renaming

bentheredonethat:
Rename ZynqMP IPI child nodes from their register addresses so the generated unit addresses match reg. Apply the corresponding child@reg naming to Versal IPI destination mailboxes during Zephyr conversion.

Preserve symbolic, alias, chosen, and other path references when renaming ZynqMP or Versal nodes. Numeric phandle references remain unchanged.

Convert legacy ZynqMP IPI identifiers to local-ipi-id and remote-ipi-id. Prune destination children to the properties accepted by the Zephyr parent child-binding. Keep zephyr,ipc on the selected ZynqMP child device and preserve Versal's zephyr,mbox-ipm adapter reference.

note this is tested for zynqmp, versal end to end and boot tested via remoteproc on target

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -128,6 +128,9 @@ def _xlnx_zephyr_convert_zynqmp_ipi_id(node):
                           value=node["xlnx,ipi-id"].value)
         node.delete("xlnx,ipi-id")
 
+    node.resolve()
+    node.sync()
+
 
 def _extra_zephyr_comp_paths(options):
     """Parse and memoize --extra-zephyr-comp arguments from options['args'].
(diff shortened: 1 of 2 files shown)


zeddii:
Thanks @bentheredonethat.

Once again your pull request made me pick up a gap in the core, these are really valuable to see. I'm not trying to waste your efforts on them, I'm just refactoring as I see things.

The need here (references surviving an IPI rename) is real, and it turned out to be a gap in core rather than something each assist should carry. I've added it to core (on master): renaming a node now follows the path references to it.

- `sdt.tree.rename(node, new_name)` renames a node in place and rewrites the path-bearing properties that referenced it — `/aliases`, `chosen`, and any path-ref value — to the new path, including references to a *descendant* of the renamed node. It uses the learned schema's classification of path-bearing properties (the same set strict output already uses to prune dangling refs), so it only touches real path-refs and catches cases an exact-string scan misses.
- `__symbols__` is already handled by core (rebuilt from each node's path on `resolve()`), so there's nothing to do there.
- It's renames only — `rename()` takes a leaf name, not a path, so it can't move a node; moves deliberately don't follow (a path-ref addresses a location; phandles follow identity).

So you can drop `_rename_node_preserve_path_refs` entirely and call it directly in its place:

```python
sdt.tree.rename(node, f"child@{node.propval('reg')[1]:x}")
...
sdt.tree.rename(remote_node, "ipi@%x" % remote_node["reg"][1])
```

The ref preservation then comes for free and is shared with every other assist. The IPI-specific parts (reg-based naming, the id conversion, child-prop pruning) stay as they are.

If you'd rather not rebase onto that immediately, the existing helper isn't wrong for the leaf cases you tested — but core `rename()` is where this belongs going forward, and it covers the descendant/edge cases the manual scan doesn't.

I'm willing to merge what you have without an update to the core API I just created, let me know.

bentheredonethat:
@zeddii  thanks for the response. I have rebased and incorporated use of the rename() method.
Please see latest commit now.

PR #804, opened 2026-08-12
lopper:assist: Generate xparameters and CMake variables from the misc…

sathishkumar-amd:
…_props node

Update Lopper to consume the misc_props node generated by SDT in pcw.dtsi and automatically propagate its properties into bare-metal xparameters.h macros and CMake metadata. This enables new user-defined properties added under misc_props to flow through to the BSP without requiring additional Lopper assist changes.

### DIFF
--- a/lopper/assists/baremetal_xparameters_xlnx.py
+++ b/lopper/assists/baremetal_xparameters_xlnx.py
@@ -576,6 +576,9 @@ def xlnx_generate_xparams(tgt_node, sdt, options):
         val = sdt.tree[tgt_node].propval('semnpi-scan', list)[0]
         plat.buf(f'\n#define XSEM_NPISCAN_EN {val}\n')
 
+    #Define for new properties added under misc_props in pcw.dtsi
+    utils.gen_misc_props_macros(plat, sdt)
+
     if is_fpd_coherent:
         plat.buf("\n#define XPAR_FPD_IS_CACHE_COHERENT \n")
     if is_lpd_coherent:
--- a/lopper/assists/bmcmake_metadata_xlnx.py
+++ b/lopper/assists/bmcmake_metadata_xlnx.py
@@ -319,6 +319,8 @@ def generate_hwtocmake_medata(sdt, node_list, src_path, repo_path_data, options,
             if sdt.tree['/'].propval('speed_grade') != ['']:
                 val = sdt.tree['/'].propval('speed_grade', list)[0]
                 fd.write(f'set(SPEED_GRADE "{val}" CACHE STRING "Speed Grade")\n')
+            #Generate cmake variable for the properties added under misc_props in pcw.dtsi
+            utils.gen_misc_props_cmake_vars(fd, sdt)
             if sdt.tree['/'].propval('board') != ['']:
                 val = sdt.tree['/'].propval('board', list)[0]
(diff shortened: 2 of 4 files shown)


sathishkumar-amd:
Hi @onkarharsh,

Please review this?

onkarharsh:
The changes look good to me.
The Integration test shows some failures which do not seem to be related to the changes done here.

zeddii:
Thanks @sathishkumar-amd — forwarding `misc_props` into xparameters + CMake so new SDT properties reach the BSP without an assist change is a nice additive approach, and it's a clean no-op for boards without a `misc_props` node. A few things before it goes in.

**1. Identifier sanitization is incomplete (the main one).** Both generators build the macro/variable name with:

```python
var_name = prop_name.replace('xlnx,', '').replace('-', '_').upper()
```

That only handles the `xlnx,foo-bar` case. Any other DT-legal name emits an invalid C/CMake identifier: a non-xlnx vendor prefix (`arm,foo` → `#define ARM,FOO`), a dot (`foo.bar` → `FOO.BAR`), or a leading digit (`2nd-x` → `2ND_X`). Since the goal is "arbitrary user-defined properties flow through without assist changes," this defeats the premise for anything outside that one naming pattern. A general scrub — replace any non-`[A-Za-z0-9_]` with `_`, and prefix a leading digit — makes it robust.

**2. `_iter_misc_props` — use the `list` cast.** It does `prop_val = node.propval(prop_name)` then `len(prop_val)`. An untyped `propval()` can return a bare scalar (not a list), on which `len()` raises. Everywhere else in lopper the pattern is `node.propval(name, list)`; using it here guarantees safe iteration for int / string / empty / list values alike.

**3. Minor — de-duplicate the name transform.** The `var_name = …` line is copy-pasted in both `gen_misc_props_macros` and `gen_misc_props_cmake_vars`; yielding the sanitized name from `_iter_misc_props` keeps one source of truth (and folds in the fix for #1). The `isinstance(val, bool)` branches are also effectively dead — DT propvals surface as int/str, not Python bool.

**4. Tests.** Nothing exercises the `misc_props` path — a small fixture with a `misc_props` node covering int, string, list, and empty properties, asserting the emitted macros/CMake vars, would lock it in and catch #1.

The red CI is the cpu-cluster-prune bug that was on master (fixed now) — a rebase clears it; onkarharsh's LGTM is informal (it doesn't set a formal review verdict). Otherwise the approach is sound.

sathishkumar-amd:
@zeddii Thanks for the review, all points are addressed in the latest push:

**1. Identifier sanitization**
  Added _misc_prop_var_name() with a general scrub: strip xlnx,/amd, prefixes, replace any non-[A-Za-z0-9_] with _, and prefix _ when the name starts with a digit. This covers vendor prefixes (arm,foo → ARM_FOO), dots (foo.bar → FOO_BAR), and leading digits (2nd-x → _2ND_X).

**2. Safe propval iteration**
  _iter_misc_props now uses node.propval(prop_name, list) so scalar values are always list-wrapped before len() / iteration.

**3. Single source of truth for name transform**
  _iter_misc_props yields (prop_name, var_name, value) with var_name from the shared helper; both gen_misc_props_macros and gen_misc_props_cmake_vars
  consume it. Removed the dead isinstance(val, bool) branches — lopper surfaces booleans as empty flags or string "true"/"false" values, which are handled explicitly.

PR #805, opened 2026-08-13
lopper: enable TTC nodes for Zephyr 

katrusatyasri:
Enable generated TTC nodes by removing the disabled status from TTC device tree nodes.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -1347,10 +1347,8 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine, options=None):
                                 child["interrupts"] = LopperProp("interrupts")
                                 child["interrupts"].value = \
                                 node["interrupts"].value[i * 4:(i + 1) * 4]
-                                child["status"] = "disabled"
                                 node.add(child)
                             node.delete("interrupts")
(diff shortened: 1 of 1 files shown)


kedareswararao (approved):

zeddii:
Thanks @katrusatyasri. Enabling the generated TTC nodes by dropping the disabled status on the parent and the counter children is right for the zephyr flow.

One thing to confirm before I take it: `xlnx_remove_unsupported_nodes` runs for the other targets too, so I want to be sure this doesn't flip TTC on for a flow (linux/baremetal) that expected them disabled. If it's effectively zephyr-only, just say so and I'll pull it in.

kedareswararao:
> Thanks @katrusatyasri. Enabling the generated TTC nodes by dropping the disabled status on the parent and the counter children is right for the zephyr flow.
> 
> One thing to confirm before I take it: `xlnx_remove_unsupported_nodes` runs for the other targets too, so I want to be sure this doesn't flip TTC on for a flow (linux/baremetal) that expected them disabled. If it's effectively zephyr-only, just say so and I'll pull it in.

@zeddii : it's zephyr only flow the reason for disabling the node earlier is to avoid build failures as there are zephyr changes that was cherry picked from upstream now those changes are merged so this pull request re enables the same.

PR #807, opened 2026-08-14
Zephyr openamp legacy mpu

bentheredonethat:
Fixups to ensure existing zephyr MPU attributes are usable for r5

### DIFF
--- a/demos/openamp/inputs/openamp-overlay-versal-2ve-2vm.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-2ve-2vm.yaml
@@ -21,7 +21,7 @@ reserved-memory:
     start: 0x985F000
     size: 0x1000
     no-map: 1
-    mpu-policy!zephyr!append: [ readable, writable, shareable, userspace ]
+    mpu-policy!zephyr!append: [ readable, writable ]
   rpu0vdev0vring0@9860000:
     start: 0x9860000
     size: 0x4000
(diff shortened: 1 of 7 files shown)


zeddii:
@bentheredonethat heads up — CI's red on the legacy sanity suite (not pytest, so it won't show up in a normal `pytest tests/` run). The failing one is "Standalone Zephyr linker and MPU generators".

Your new `_memory_attributes` is rejecting `/axi/sram@0`:

> zephyr_mpu: /axi/sram@0: non-static R-profile memory must be privileged read/write and execute-never

the assist returns false and the test fails — the RW + execute-never rule is stricter than what that sram case provides.

To reproduce locally (pytest alone won't catch it):

```
./scripts/run_tests.sh          # this is what CI runs
# or just the sanity suite:
python3 lopper_sanity.py --all
```

Failing run: https://github.com/devicetree-org/lopper/actions/runs/31827622724

Green that and I'll dig into the review.

bentheredonethat:
ok @zeddii  updated

zeddii:
@bentheredonethat closer — `sram@0` is happy now and "Standalone Zephyr linker and MPU generators" passes. But it moved: "OpenAMP Zephyr R5 demo YAML policies" now fails on the same rule, this time on a trace buffer:

> zephyr_mpu: /reserved-memory/trace@985f000: non-static R-profile memory must be privileged read/write

Are you seeing that one locally too? Both 3.12 and 3.13 hit it, and it's the only red left.

Latest run: https://github.com/devicetree-org/lopper/actions/runs/31841132570

bentheredonethat:
@zeddii  with latest it passes all now -- you see this too, yes?

zeddii:
@bentheredonethat yep, green here too. That `trace@985f000` error was on `b44c9af` — you'd already re-pushed `b2d460d75` by the time my comment landed, so I was staring at the older head. Latest passes clean on both 3.12 and 3.13. Good to merge.

PR #808, opened 2026-08-17
gen_domain: Fix up TTC generation

bentheredonethat:
Some child properties are missing so ensure they are present for zephyr builds

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -1348,33 +1348,30 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine, options=None):
                         if 'psv_cortexr5' in machine or 'psu_cortexr5' in machine:
                             node["compatible"] = "xlnx,ttcps"
                             if node.propval('interrupt-names') == ['']:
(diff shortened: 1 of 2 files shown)


bentheredonethat:
@kedareswararao  -- note there were some issues with builds for versal2 - fixups here for ttc

bentheredonethat:
@zeddii  this is needed for latest gen domains for zephyr AFAIK

katrusatyasri:
Hi @bentheredonethat,
The upstream TTC counter driver has already been merged into the internal Zephyr tree, and the corresponding lopper changes required for the new TTC devicetree hierarchy have also been merged.

This change regenerates the legacy TTC node format (counter0@<base>, counter1@<base>, etc.) using the timer-id property. The upstream TTC driver does not use this format and expects the newer hierarchy with an xlnx,ttc parent node and counter@0/1/2 child nodes.

The legacy timer-id based node generation is no longer needed and should not be reintroduced. We should keep the current upstream-aligned TTC hierarchy generated by the existing lopper changes.

kedareswararao:
@bentheredonethat : Please close the pull request as the existing code is already generating the required nodes for the latest driver if you still see issues please sync with @katrusatyasri

PR #809, opened 2026-08-19
lops: add lop-prune-domains to strip /domains from final DTS

Rajukumar45:
Add a standalone modify lop that deletes /domains ("/domains::"). It is meant to run as the final lop in a chain, after gen_domain_dts and any domain-consuming assists, and is a harmless no-op when no /domains node is present. Safe to append to the linux_dt, zephyr_dt, and baremetal domain flows.

### DIFF
--- a/lopper/lops/lop-prune-domains.dts
+++ b/lopper/lops/lop-prune-domains.dts
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2026 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/dts-v1/;
+
+/ {
+        compatible = "system-device-tree-v1";
+        lops {
+                lop_0 {
+                        // Delete the /domains node so it does not leak into the final DTS.
+                        // Done via a code lop so it is a no-op when /domains is absent.
+                        compatible = "system-device-tree-v1,lop,code-v1";
+                        code = "
+                            try:
+                                domains_node = tree['/domains']
+                            except:
+                                domains_node = None
+
+                            if domains_node:
+                                tree.delete(domains_node)
+                                tree.sync()
+                        ";
+                };
+        };
+};


kedareswararao (approved):

zeddii:
Thanks @Rajukumar45 — stripping `/domains` from the final DTS is a reasonable thing to want a lop for.

There's no need for a `code-v1` lop with a try/except here, though. Lopper already has a node-delete primitive: a `modify` lop whose operand is a path with an empty property and value, `modify = "/domains::";`. See `lop-delete-chosen.dts` for the canonical form:

```dts
lop_1 {
        compatible = "system-device-tree-v1,lop,modify";
        modify = "/chosen::";
};
```

The selector on the left of a `modify` operand is match-or-skip by design: if the path matches a node it's operated on, and if it matches nothing the operation is simply skipped. So a plain `modify = "/domains::";` already gives you the "no-op when `/domains` is absent" behaviour your comment is reaching for — the try/except and the manual `tree.delete()`/`tree.sync()` are doing by hand what the primitive does natively.

One caveat: that tolerance had regressed on master — a node delete against an absent path was hard-exiting rather than skipping. That's now fixed (commit `e1de802c`, "tree: make a node-delete modify operand tolerant of a missing selector"), so the primitive behaves as intended again. Please rebase onto current master and the simpler form below will work:

```dts
/dts-v1/;

/ {
        compatible = "system-device-tree-v1";
        lops {
                lop_0 {
                        compatible = "system-device-tree-v1,lop,modify";
                        modify = "/domains::";
                };
        };
};
```

This keeps the lop declarative and consistent with the other delete lops in the tree.

Rajukumar45:
Hi @zeddii 

You are correct. This can be implemented using modify, and I have already tried that approach. However, using modify fails when the DTS file does not contain a domains node. That is why I am using code-v1 instead. Rather than terminating with an error in such cases, I added a check in the LOP file so the flow exits gracefully when the domains node is absent.

If the latest commit fixes above issue sure, I can rework and send v2.

PR #810, opened 2026-08-20
lops: add lop-prune-domains to strip /domains from final DTS

Rajukumar45:
Add a standalone modify lop that deletes /domains ("/domains::"). It is meant to run as the final lop in a chain, after gen_domain_dts and any domain-consuming assists.


zeddii:
Looks good, and thanks for the quick turn around. The bug that you hit when first trying that was a regression, I'm glad to have found it as part of this.

PR #813, opened 2026-08-21
openamp: select domain CPUs from the SDT mask

bentheredonethat:
Decode the CPU mask in the standard cpus tuple when matching OpenAMP domains and reporting processors. Retain cluster_cpu as a fallback for legacy inputs.

Derive RPU metadata from the selected CPU during YAML expansion and accept both integer and string CPU masks.

### DIFF
--- a/demos/openamp/inputs/libmetal-overlay-zynqmp.yaml
+++ b/demos/openamp/inputs/libmetal-overlay-zynqmp.yaml
@@ -66,7 +66,7 @@ domains:
     compatible: openamp,domain-v1
     cpus:
       - cluster: cpus_r5_1
-        cpumask: 0x2
+        cpumask: 0x1
         mode:
           secure: true
           el: 0x3
--- a/demos/openamp/inputs/openamp-overlay-versal-2ve-2vm.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-2ve-2vm.yaml
@@ -425,7 +425,7 @@ domains:
     cpus:
       - cluster: cpus_r52_1
         cluster_cpu: cortexr52_1
-        cpumask: 0x2
+        cpumask: 0x1
         mode:
           secure: true
           el: 0x3  
@@ -479,7 +479,7 @@ domains:
     compatible: openamp,domain-v1
     cpus:
       - cluster: cpus_r52_8
-        cluster_cpu: cortexr52_9
+        cluster_cpu: cortexr52_8
         cpumask: 0x1
         mode:
           secure: true
--- a/demos/openamp/inputs/openamp-overlay-versal-lockstep.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-lockstep.yaml
@@ -113,7 +113,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r5_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode:
         access:
             - <<+: *openamp_channel_1_access_srams
--- a/demos/openamp/inputs/openamp-overlay-versal-net-split.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-net-split.yaml
@@ -113,7 +113,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r52_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode: 0x1
               secure: true
         access:
@@ -168,5 +168,4 @@ domains:
                      - rpu0vdev0vring0
                      - rpu0vdev0vring1
                      - rpu0vdev0buffer
-
 
--- a/demos/openamp/inputs/openamp-overlay-versal-net.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-net.yaml
@@ -113,7 +113,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r52_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode: 0x1
               secure: true
         access:
@@ -168,5 +168,4 @@ domains:
                      - rpu0vdev0vring0
                      - rpu0vdev0vring1
                      - rpu0vdev0buffer
-
 
--- a/demos/openamp/inputs/openamp-overlay-versal-split.yaml
+++ b/demos/openamp/inputs/openamp-overlay-versal-split.yaml
@@ -109,7 +109,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r5_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode: 0x1
               secure: true
         access:
--- a/demos/openamp/inputs/openamp-overlay-zynqmp-dev-mem.yaml
+++ b/demos/openamp/inputs/openamp-overlay-zynqmp-dev-mem.yaml
@@ -119,7 +119,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r5_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode: 0x1
               secure: true
         access:
--- a/demos/openamp/inputs/openamp-overlay-zynqmp-lockstep.yaml
+++ b/demos/openamp/inputs/openamp-overlay-zynqmp-lockstep.yaml
@@ -113,7 +113,7 @@ domains:
             - "openamp,domain-v1"
         cpus:
             - cluster: cpus_r5_1
-              cpumask: 0x2
+              cpumask: 0x1
               mode:
         access:
             - <<+: *openamp_channel_1_access_srams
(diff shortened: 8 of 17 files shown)


bentheredonethat:
@zeddii  please take a look 

Lopper currently uses non-standard cluster_cpu key instead of the SDT-standard cpu_mask bitfield to identify CPU cores assigned to a domain for openamp - fix this here

preserve backward compatibility but also enable openamp plugins and yaml translation to work purely off cpu mask if able

zeddii:
I'm traveling at the moment, but will have a closer look at this ASAP. This one, I do need to look into more and run some of my own tests.

zeddii:
Ben, here's a mix of AI review and my own digging, anything from AI is
noted. I checked AI findings all myself and they match what I was
thinking, I just didn't have time to do all the digging myself.

I went looking at what this does downstream of the OpenAMP assists,
since the mask feeds `domain_access` too. Two things worth raising
before it is mergeable. One is an impact you may not have measured,
the other is a genuine ambiguity that should be settled rather than
have each path answer separately.

## The mask fix reaches much further than OpenAMP

`_cpu_mask_value` is the highest-impact change here. Because `int(mask, 16)` threw on
YAML-native integers and the bare `except` zeroed it, *every* unquoted mask was landing
in the DTS as `0`. Expanding `openamp-overlay-versal-2ve-2vm.yaml`:

```
before                              after
cpus = <&cpus_a78    0x0 0x3>;      cpus = <&cpus_a78    0xf 0x3>;
cpus = <&cpus_r52_0  0x0 ...>;      cpus = <&cpus_r52_0  0x1 ...>;
cpus = <&cpus_r52_1  0x0 ...>;      cpus = <&cpus_r52_1  0x2 ...>;
cpus = <&cpus_r52_6  0x0 ...>;      cpus = <&cpus_r52_6  0x1 ...>;
```

The consequence outside OpenAMP: `lopper_lib.cpu_refs` (called from
`domain_access.py:546`) gates all of its pruning on the mask having selected something:

```python
refd_cpus, unrefd_cpus = lopper_lib.cpu_refs( sdt.tree, cpu_prop, verbose )
if refd_cpus and device_prune:
    ...cluster-level filter, then delete unrefd_cpus...
```

With `mask == 0` that block never runs, so YAML-expanded domains have been getting no
CPU pruning at all: a superset, which is why nobody hit it. (Domains written directly
in DTS carry real masks and have pruned correctly for years; that path is untouched.)

So this doesn't adjust pruning, it **switches it on** for every YAML-expanded domain at
once, cluster-level and intra-cluster both. `domain_access -t /domains/APU_Linux` on
the versal2 SDT:

| | a78 cores | r52 cores | output |
|--------|-----------|-----------|--------|
| before | 8 | 10 | 303 KB |
| after | 4 (`0xf`) | 0 | 238 KB |

The new numbers are the correct ones. But it's a project-wide activation riding inside
an OpenAMP patch, so I'd rather see `_cpu_mask_value` split into its own commit that
can be called out in isolation.

The other side effect: because the masks never took effect, they've never been
validated. Which leads to the real question.

## The big question: What is the mask indexed by?

`chapter3-execution-domains.rst:150` says the bitfield covers "the subset of CPUs **in
the cluster**", and `cpu_refs` implements that positionally over the cluster's CPU
children:

```python
sub_cpus = cpu_node.subnodes( children_only=True, name="cpu@.*" )
if check_bit_set( cpu_mask, idx ):
    sub_cpu_node = sub_cpus[idx]
```

An AI driven check of the code follows to generate the table and indexing discussion:

`_openamp_cpu_index` here indexes by `reg` instead. That leaves two independent decoders
of the same property: `cpu_refs`, reached from `domain_access.py:546` when a domain is
scoped, and `_openamp_domain_selects_cpu`, reached from the four call sites in
`openamp_xlnx.py` (127, 209, 338, 1383). With masks now live, the same
`cpus = <&cluster mask level>` names different cores depending on which one reads it:

| domain | cluster | mask | `cpu_refs` (positional) | this patch (`reg`) |
|------------|------------|------|--------------------------------------|--------------------|
| APU_Linux | cpus_a78 | 0xf | cpu@0, cpu@100, cpu@10000, cpu@10100 | cpu@0 |
| RPU_Zephyr | cpus_r52_0 | 0x1 | cpu@0 | cpu@0 |
| RPU_BM | cpus_r52_1 | 0x2 | (none) | cpu@1 |
| RPU_ISP0 | cpus_r52_6 | 0x1 | cpu@6 | (none) |
| RPU_ISP3 | cpus_r52_9 | 0x1 | cpu@9 | (none) |

Six of seven disagree, and in both directions: `RPU_BM` only resolves under `reg`, the
ISP domains only under positional. The in-tree YAML has been written both ways, which
is what you'd expect from values that never took effect.

One constraint on the answer: `reg` can't be the index for A-profile clusters, where it
carries MPIDR affinity bits rather than an ordinal

```
cortexa78_0  cpu@0       reg=0x0       -> bit 0
cortexa78_1  cpu@100     reg=0x100     -> bit 256
cortexa78_7  cpu@30100   reg=0x30100   -> bit 196864
```

so only core 0 is reachable in a 32-bit mask. That's the `APU_Linux` row above going
from 4 cores to 1.

There's also a third opinion in the tree, and it's the one that *writes* these values.
`sdt_devices` builds cpumasks in `_devices_core.py` from the `cpu-map` linear index
where one exists, and sequential enumeration otherwise. Run against the versal2 SDT it
emits:

```
cpus_a78:    cpumask: 0xff
cpus_r52_1:  cpumask: 0x1
```

Both cluster-relative. So a mask lopper generates itself doesn't survive a round trip
through `_openamp_cpu_index`: that `0xff` comes back as `cortexa78_0` alone. It also
emits `0x1` for `cpus_r52_1` where the overlay says `0x2`, which is independent
confirmation that those hand-written masks are wrong rather than a second convention.

AI block done ...

My read is that the spec is cluster-relative, and that `cpu_refs` has it right: the
chapter 3 wording is explicit about the subset being "in the cluster", the generator
already writes them that way, and the MPIDR case rules out `reg` as a workable index
rather than merely making it a second option.
So I think `_openamp_cpu_index` should select positionally over the cluster's CPU
children, and that both decoders should call one shared helper so they can't drift
again.

More AI driven feedback here:

Taking that position breaks some of your inputs: it makes `RPU_BM` and the
ZynqMP `r5_1` domains wrong data. A mask of `0x2` against a single-child cluster
selects nothing cluster-relative, so those masks need correcting to `0x1` alongside the
change. That currently fails safe, since an empty `refd_cpus` skips pruning, but it
should be fixed rather than relied on. 

AI done

If you read the spec differently, that's worth resolving before either
path starts honouring the mask, because right now they'd honour it
differently.

## Smaller things (AI findings)

`cpumask: 0x0` in `openamp-overlay-zynqmp-split.yaml` (the a53 host) and
`openamp-overlay-versal-net-lockstep.yaml`: the guard
`if cluster is not None and mask is not None` is satisfied by `0`, so it takes the
strict path, matches nothing, and doesn't fall through to `cluster_cpu`. Treating `0`
as unspecified would be safer.

`RPU_ISP2` has `cluster: cpus_r52_8` with `cluster_cpu: cortexr52_9`: looks like a
copy-paste from the ISP3 block. Pre-existing, but the new fallback resolves the label
rather than the cluster, so it now reaches `core_num`. Cheap to warn when the two point
at different cores, which would have surfaced it.

Test-wise, the new cases put `cpu@0` and `cpu@1` in one cluster, which is the shape
where both index conventions agree. A single-core cluster at a non-zero unit address,
and an A-profile cluster with MPIDR `reg`, would pin the behaviour down.

AI findings done

Happy to take the mask-audit side of this if it helps split the work up.

zeddii:
Thanks for turning that around so quickly, and for splitting the mask fix out into its
own commit. The shared `cpu_nodes_from_mask()` in `lopper_lib.py` is exactly the shape
I was hoping for: one definition of what a cpumask means, with both the domain scoping
and the OpenAMP path going through it.

I re-ran things against the rewrite. AI assisted with the runs, I checked the results.

The emitted triplets are now self-consistent, and they agree with what `sdt_devices`
generates for the same SDT, so the round trip closes:

```
cpus = <&cpus_a78    0xf 0x3>;
cpus = <&cpus_r52_0  0x1 0x80000003>;
cpus = <&cpus_r52_1  0x1 0x80000003>;
cpus = <&cpus_r52_6  0x1 0x80000003>;
```

The three problems I raised are gone:

    RPU_ISP0/1/2/3 and RPU_BM      all resolve now (were NO MATCH)
    domain_access -t APU_Linux     4 a78 cores, 0 r52 (correct for 0xf)
    cpumask 0x0                    falls back and matches again

Changing the guard to `if cluster is not None and mask:` is a better answer for the
zero masks than editing the two overlays, since it fixes any input rather than the two
we happen to ship.

Test suites are green here: 1103 passed / 4 skipped under pytest, and 124 passed / 0
failed in the sanity suite.

One thing left before this can merge: DCO is failing because all three commits lost
their `Signed-off-by` in the force-push. A `git rebase --signoff` over the three and a
re-push should clear it. Both test jobs (3.12 and 3.13) are already passing, so that
check is the only thing holding it.

The one suggestion I made that isn't in here is warning when `cluster_cpu` and the mask
point at different cores. That is moot for the current inputs now that the
`cortexr52_8` typo is fixed, but it is the guard that would have caught it, and it
would catch the next one. Fine as a follow-up rather than something to fold in here.

Bruce

PR #820, opened 2026-08-31
Openamp timer handling

bentheredonethat:


lopper/assists/openamp_xlnx.py
```diff
@@ -93,63 +93,77 @@ def xlnx_openamp_keep_node(linux_dt, zephyr_dt, node, tree):
...
-    if target_os == "linux_dt":
-        domains = [ n for n in tree["/domains"].subnodes(children_only=True) if n.propval("os,type") == [target_os.replace("_dt","")] ]
-    else:
-        # otherwise filter by machine
-        match_cpunode = get_cpu_node(sdt, {'args':[machine]})
-        domains = [ n for n in tree["/domains"].subnodes(children_only=True) if match_cpunode.parent == sdt.tree.pnode(n.parent.parent.propval("cpus")[0]) ]
-
-    if not domains:
+    if target_os != "baremetal_dt":
         return False
```
zeddii: This is the more focused area that I'd like confirmed. The old code reached the `cdns,ttc` restore via an *else*. i.e. every `dt_type` that wasn't `linux_dt`. This requires the exact string `"baremetal_dt"`.

`dt_type` is set three different ways in `parse_openamp_args`:
- the 2-arg form → `"linux_dt"` / `"zephyr_dt"`
- the 1-arg form → `"baremetal_dt"`
- **the argparse form → `config["dt_type"] = config["os"]`**, i.e. whatever `--os` says, or `None` when `--os` is omitted

That third path does reach this function — only `--openamp_output_filename` returns early, so e.g. a `--libmetal_output_file` run with `--os freertos` (or no `--os` at all) lands here and now no-ops where it previously restored the binding. The comment you removed named this case explicitly: *"baremetal / freertos / zephyr case"*.

Nothing in-tree passes `--os`, so CI can't catch it — it'd be the external generators. Is FreeRTOS confirmed not to need this any more, or should the guard be an allow-list of the OS-flavoured values rather than the one literal? Or is this just an impossible/completely invalid case ?

lopper/assists/openamp_xlnx.py
```diff
@@ -93,63 +93,77 @@ def xlnx_openamp_keep_node(linux_dt, zephyr_dt, node, tree):
...
-            elif "uio" in node.propval("compatible"):
-                node["status"] = "okay"
+    if not relation_groups:
+        return False
 
-        return True
+    # A relation's timer property names only the timer assigned to that
+    # OpenAMP/Libmetal channel.  It is not an allow-list for all system timers.
+    timer_phandles = []
+    for relation in relation_groups[0].subnodes(children_only=True):
```
zeddii: `relation_groups[0]` reads only the **first** relation-type container. If a domain carries both an `openamp-relation` and a `libmetal-relation` under `domain-to-domain`, the timers named by the second one never get restored.

The old code walked `domains[0].subnodes(children_only=True)`, which — by the descendant-walk behaviour you document above at line 125 — swept every relation under the domain and so saw both containers.

The new test builds a single container, so it wouldn't catch it either. Is one container per domain guaranteed here? If not, iterating all of `relation_groups` looks like it'd be a one-line change.

lopper/assists/openamp_xlnx.py
```diff
@@ -93,63 +93,77 @@ def xlnx_openamp_keep_node(linux_dt, zephyr_dt, node, tree):
...
-    # this whole routine can go away. Until then this is a holdover
-    # to ensure that UIO / linux is there if needed.
-    for n in tree["/axi"].subnodes(children_only=True, name="timer@*"):
-        n["compatible"] = "cdns,ttc"
+    # The shared source tree may contain Linux's UIO override.  Bare-metal
+    # consumers use the native TTC binding, so undo that override only for the
+    # timers explicitly named by their relation.  Do not touch other timers.
+    for timer_phandle in timer_phandles:
+        timer_node = tree.pnode(timer_phandle)
+        is_linux_uio = (timer_node is not None and "uio" in timer_node.propval("compatible"))
```
zeddii: Minor consistency: `propval("compatible")` without the `list` arg, where `xlnx_openamp_keep_node` at line 88 uses `propval('compatible', list)` for the same test. Both happen to work for a single-string `compatible`, but the unqualified form falls back to a substring match if the value ever comes back as a bare string rather than a list. Worth matching line 88.

lopper/assists/openamp_xlnx.py
```diff
@@ -93,63 +93,77 @@ def xlnx_openamp_keep_node(linux_dt, zephyr_dt, node, tree):
...
         True if success. Else False.
 
     Algorithm:
-        Find domains matching target OS.
-        For each relation, look for timer property.
-        Based on timer property trim irrelevant timers.
+        Linux-compatible UIO properties come from the domain YAML and Zephyr
+        does not use the Linux ``cdns,ttc`` binding.  Only bare-metal output
+        needs the relation-selected timer restored after a global Linux UIO
+        overlay has been applied.
```
zeddii: This premise is the important one for dropping the whole Linux branch, and it's the part I can't verify. The old Linux path did two things this no longer does: it **deleted** `/axi/timer@*` nodes not named by a relation, and it forced `status = "okay"` on the UIO ones that were.

The new tests confirm the function no-ops for `linux_dt`, which is the new intent ... but nothing demonstrates the domain YAML actually supplies the `status` that used to be set here. Could you point at the YAML (or the assist) that provides it now? If it does, this is fine and I'd suggest naming it in the commit message; if it doesn't, Linux output loses an enabled UIO timer and that'd be a quiet one to debug later.

zeddii (commented):
Thanks @bentheredonethat. Narrowing this from "rewrite every timer" to "restore the ones a relation actually names" is the right direction, and preserving unrelated TTC providers like the Kria PWM fan is a real fix. The tests are a good addition too.

Two things I'd like confirmed before this goes in: both are areas where things are more focused than the old code. I commented inline with the specific questions:

1. **Non-`baremetal_dt` callers.** The old else-branch caught every dt_type that wasn't `linux_dt`; this requires the literal string `"baremetal_dt"`. `dt_type` can also arrive as a free-form `--os` value (or `None`), and that path does reach the call site. The code you're replacing described itself as the "baremetal / freertos / zephyr case".
2. **Only the first relation container is read.** A domain carrying both an OpenAMP and a Libmetal container will only have one of them consulted.

If you've already checked both, a one-line answer on each is enough for me and I'll take it. Also curious where the Linux `status = "okay"` now comes from (inline) — the old code set it here, and I'd like to know what supplies it instead.

bentheredonethat:
Hi @zeddii  thanks for the review -- 

these checks found two concrete gaps and one unsupported assumption. 

I’ll iterate all relation containers and use the list form for compatible. 

he ZynqMP overlay changes the selected timer to uio, but the base TTC can be disabled in the input SDT, so I’ll retain relation-scoped Linux enablement without deleting unrelated timers. 

For non-Linux targets, I’ll ensure  the FreeRTOS/legacy path has the compatible string preserved

bentheredonethat:
Hi @zeddii -- to clarify:
- for linux - if timer is specified and UIO then yes make sure its enabled (status okay)
- else - dont modify it - the input SDT and other plugins' processing can modify it but there is not a reason to do so from openamp case -- even in libmetal baremetal rpu1 case or legacy freertos openamp case -- the input SDT is already correct and this handling is not needed 
given the above i want to reduce the scope of this routine -- linux only, if timer specified for uio then yes make sure its enabled -- nothing else

bentheredonethat:
Bare-metal, FreeRTOS, Zephyr, and unspecified targets remain unchanged

lopper/assists/openamp_xlnx.py
bentheredonethat: the point is that ONLY in linux case - IF TTC node has UIO - then make sure its set to OK as user is explicitly mentioning it - i dont want this to depend on domain access

alternatively we could update libmetal domain yaml here to also set status to OK I guess 

see below for current YAML
`the current impl is not no-op it is op to set UIO specified linux TTCs to status OK
YAML reference below 


```
reserved-memory:
  ranges: true
  "#size-cells": 2
  "#address-cells": 2

  libmetal_elf@9968000:
    label: libmetal_elf
    start: 0x9968000
    size: 0x60000
    no-map: 1
  libmetal_desc0@99c8000:
    label: libmetal_desc0
    start: 0x99c8000
    size: 0x4000
    no-map: 1
  libmetal_desc1@99cc000:
    label: libmetal_desc1
    start: 0x99cc000
    size: 0x4000
    no-map: 1
  libmetal_data@99d0000:
    label: libmetal_data
    start: 0x99d0000
    size: 0x40000
    no-map: 1

axi:
  libmetal_uio_desc0@99c8000:
    linux,uio-name: "libmetal-desc0"
    start: 0x99c8000
    size: 0x4000
    compatible: "uio"
  libmetal_uio_desc1@99cc000:
    linux,uio-name: "libmetal-desc1"
    start: 0x99cc000
    size: 0x4000
    compatible: "uio"
  libmetal_uio_data@99d0000:
    linux,uio-name: "libmetal-data"
    start: 0x99d0000
    size: 0x40000
    compatible: "uio"

  timer@ff130000:
    compatible: "uio"
    linux,uio-name: "libmetal-timer"
  # linux -> rpu1 libmetal IPI
  ipi@ff350000:
    compatible: "uio"
    linux,uio-name: "libmetal-ipi"

domains:
  APU_Linux: # host in channel from a53-0
    compatible: openamp,domain-v1
    domain-to-domain:
      compatible: openamp,domain-to-domain-v1
      libmetal-relation:
        compatible: libmetal,ipc-v1
        relation0:
          remote: R5_1_BAREMETAL
          carveouts: [ libmetal_uio_desc0@99c8000, libmetal_uio_desc1@99cc000, libmetal_uio_data@99d0000 ]
          mbox: ipi_8_to_ipi_2
          timer: ttc2

  R5_1_BAREMETAL:
    compatible: openamp,domain-v1
    domain-to-domain:
      compatible: openamp,domain-to-domain-v1
      libmetal-relation:
        compatible: libmetal,ipc-v1
        relation0:
          host: APU_Linux
          carveouts: [ libmetal_uio_desc0@99c8000, libmetal_uio_desc1@99cc000, libmetal_uio_data@99d0000 ]
          mbox: ipi_2_to_ipi_8 #IPI2 Libmetal example app (RPU)
          timer: ttc2
          elfload: [ libmetal_elf@9968000 ]

    sram:
      - dev: libmetal_elf@9968000
        start: 0x9968000
        size: 0x60000
      - dev: psu_r5_1_atcm_global@ffe90000
        size: 64K
        start: 0xffe90000
````

lopper/assists/openamp_xlnx.py
bentheredonethat: addressed now

lopper/assists/openamp_xlnx.py
bentheredonethat: ok now:

- freertos / baremetal / zephyr / other legacy case -- ensure pre-existing compatible string is used -- use SDT's provided status as is
- linux - if TTC is UIO - then ensure compatible and status are proper

Confirmed FreeRTOS and Baremetal do not need  UIO compatible or deletion

lopper/assists/openamp_xlnx.py
bentheredonethat: This is now addressed with latest:

`166             and "uio" in timer_node.propval("compatible", list)`

PR #821, opened 2026-09-01
lopper: assists: Propagate xiltimer into third-party OS lib_list buckets

harishpamarthi:
Extend baremetal_getsupported_comp_xlnx to ensure xiltimer is present in
  lib_list.yaml for third-party OS buckets (e.g. ucos), not only for
  standalone and FreeRTOS.
  Third-party OS BSPs depend on xiltimer, but xiltimer's YAML metadata
  lists supported_os as standalone and FreeRTOS only. During lib discovery,
  xiltimer is therefore registered under the standalone bucket and is not
  automatically added to third-party OS buckets created from .repo.yaml.
  When writing lib_list.yaml, copy the standalone xiltimer entry into each
  non-native OS bucket (all os_bsp_keys except standalone and freertos) if
  that bucket does not already contain xiltimer.
  This avoids modifying xiltimer.yaml or adding pyesw-side library merge
  logic, and keeps third-party OS lib_list.yaml consistent with standalone
  BSP expectations.

### DIFF
--- a/lopper/assists/zephyr_supported_comp.yaml
+++ b/lopper/assists/zephyr_supported_comp.yaml
@@ -6,6 +6,12 @@ amd,mbv32:
     - riscv,isa
     - clock-frequency
 
+xlnx,pmcl-trng-11.0:
+  required:
+    - compatible
+    - status
+    - reg
+
 xlnx,xps-intc-1.00.a:
   required:
     - compatible


zeddii (commented):
The logic here is correct as far as I can tell. I checked the two things that looked like they might be an issue, and neither seems to be one: `supported_libs_dict[proc_name]["standalone"]` is always present (`get_os_bsp_keys` seeds `["standalone", "freertos"]`, and the dict is built from those keys), and sharing one entry object across buckets produces no YAML anchors because `VerboseSafeDumper.ignore_aliases` returns `True`. So this is two small suggestions.

The one I'd actually like changed is the placement (see my inline comment). The other is a style point you can take or leave.

No test on this one either. A build-system-facing artifact like `lib_list.yaml` changing is the sort of thing worth pinning, even loosely.

lopper/assists/baremetal_getsupported_comp_xlnx.py
```diff
@@ -178,6 +178,16 @@ def xlnx_baremetal_getsupported_comp(tgt_node, sdt, options):
...
 
     with open(os.path.join(sdt.outdir, 'lib_list.yaml'), 'w') as fd:
+        #Add xiltimer as a dependency to all thirdparty OS entries.
+        NATIVE_OS = frozenset(("standalone", "freertos"))
+        xiltimer_entry = supported_libs_dict[proc_name]["standalone"].get("xiltimer")
+        if xiltimer_entry:
+            for os_key in os_bsp_keys:
+                if os_key not in NATIVE_OS:
+                    bucket = supported_libs_dict[proc_name][os_key]
+                    if "xiltimer" not in bucket:
```
zeddii: This block builds up a dict, which has nothing to do with the file handle it's nested inside — and `open(..., 'w')` has already truncated `lib_list.yaml` by the time it runs. If anything in here raised, you'd be left with a zero-length `lib_list.yaml` on disk rather than the previous good one.

I don't think anything here *can* raise today ... every key in `os_bsp_keys` is seeded into `supported_libs_dict[proc_name]` at construction, so the lookups are all safe. So this is robustness rather than a live bug. But the fix is to move the block above the `with`, which costs nothing and keeps the failure window off the truncated file:

```python
# Add xiltimer as a dependency to all thirdparty OS entries.
NATIVE_OS = frozenset(("standalone", "freertos"))
xiltimer_entry = supported_libs_dict[proc_name]["standalone"].get("xiltimer")
if xiltimer_entry:
    for os_key in os_bsp_keys:
        if os_key not in NATIVE_OS:
            bucket = supported_libs_dict[proc_name][os_key]
            if "xiltimer" not in bucket:
                bucket["xiltimer"] = xiltimer_entry

with open(os.path.join(sdt.outdir, 'lib_list.yaml'), 'w') as fd:
    fd.write(yaml.dump(supported_libs_dict, ...))
```

It also reads better ... the `with` then does one thing.

lopper/assists/baremetal_getsupported_comp_xlnx.py
```diff
@@ -178,6 +178,16 @@ def xlnx_baremetal_getsupported_comp(tgt_node, sdt, options):
...
+                        bucket["xiltimer"] = xiltimer_entry
```
zeddii: Style, take or leave: this puts the *same* dict object into every third-party bucket alongside the standalone one, so a later mutation through any one of them would silently change all of them.

It's harmless as the code stands as nothing mutates these after this point, and the dumper has `ignore_aliases` on so the emitted YAML is identical either way. But `dict(xiltimer_entry)` costs nothing and means the buckets are independent, which is what the surrounding code reads as if it's doing.

PR #830, opened 2026-09-04
Generate cortex r52 tcm configuration

bentheredonethat:
@zeddii  this is needed for cleaner RPU ELF code execution on ARM R52s

### DIFF
--- a/lopper/assists/zephyr_linker.py
+++ b/lopper/assists/zephyr_linker.py
@@ -15,6 +15,8 @@ from zephyr_memory import (
     LayoutError, parse_layout, zephyr_argument_parser,
 )
 
+TCM_REGION_ENABLE = 0x1
+
 
 def is_compat(node, compat_string_to_test):
     """Identify the assist compatibility string.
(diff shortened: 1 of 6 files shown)


bentheredonethat:
I re-opened as I was waiting for the corresponding zephyr TCM patch to land

zeddii (approved):
All five points addressed, with tests. `test_tcm_config_declares_the_zephyr_startup_symbols` pins the symbol names, which is the cross-repository contract that would otherwise rot silently.

Deleting the duplicated validation in `_infer_profile` rather than syncing a second copy of the constant was the better call.

I also settled the ATCM question I couldn't verify last time. `zephyrproject-rtos/zephyr#118037` consumes the same `z_arm_tcm_*_region` names, and its reset path does `tst r1, #1` / `beq` ahead of the `mcr` — a clear bit 0 skips the write, so the zero word preserves the boot-firmware configuration rather than disabling the bank. The symbols are `.weak` there, so a script without them still links.

One leftover, non-blocking: `zephyr_memory.py:600`'s `ATCM must use local address 0x0` is the same dead check as the block you removed, since `_normalized_memory` has already pinned ATCM to `0x0` before `_infer_profile` sees it. Worth taking next time you're in the file.

PR #833, opened 2026-09-08
Openamp power domains with latest

bentheredonethat:
make sure openamp workflows are working with latest system device trees - regarding power-domains property update

also harden err handling for such cases
and add sanity to left shift catching SDT repo changes that cause breakage

### DIFF
--- a/lopper/assists/openamp_xlnx.py
+++ b/lopper/assists/openamp_xlnx.py
@@ -1682,9 +1682,13 @@ def xlnx_openamp_parse(sdt, options, verbose = 0 ):
     Returns:
         bool: True when processing succeeds or no domains exist, False on errors.
 
+    Raises:
+        SystemExit: If a requested OpenAMP relation cannot be processed.
+
     Algorithm:
         Parses assist arguments, checks for OpenAMP-compatible domains, delegates
-        relation handling when appropriate.
+        relation handling when appropriate. Relation-processing failures are
+        fatal because continuing would write an incomplete OpenAMP device tree.
     """
     # Xilinx OpenAMP subroutine to parse OpenAMP Channel
     # information and generate Device Tree information.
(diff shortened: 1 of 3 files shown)


zeddii (commented):
Confirmed the changes and the tests cover them. One thing I'd like in before this merges, then I'm happy with it.

Please capture this in a comment: the direct fallback (accepting `pd_id` when it is already a legacy key) is only safe because the two key spaces don't overlap — SCMI is `0x44`–`0x61`, legacy is `0xf`–`0x12` plus the `0x1831xxxx` block. If a future SCMI ID ever lands in the legacy range, that path silently selects the wrong address descriptor instead of erroring. Worth stating next to the table, since nothing enforces it.

Two notes, neither blocking:

- Renaming `memory_nodes` to `legacy_memory_nodes` is right, and the only in-tree consumer is updated here. It is reachable through `import *` though, so anything out-of-tree referencing it breaks on upgrade — release-note material rather than a code change.
- The relation failure becoming fatal matches what you did for the Libmetal output path. It does mean a build that previously completed with a warning and no remoteproc nodes now fails outright. That is the point, but it will surface as a new failure for anyone who was unknowingly relying on the old behaviour.

The 18 SCMI ID values themselves I can't check from here — the table resolves cleanly, but the IDs rest on your hardware knowledge.

bentheredonethat:
@zeddii  thanks for review!! i have added comments and updated commit messages to answer your Qs and comments - thanks again

zeddii (approved):
All in — the comment covers it, and the "reassess whenever SCMI IDs are extended" line is a better ending than what I asked for, since it tells the next person what to do rather than just what is true. Putting the rename caveat in the commit message is the right home for it too.

Re-checked the table on the new head: all 18 targets resolve, no key-space overlap. Good to merge.

PR #834, opened 2026-09-09
lopper: assists: restore misc_props pruning after Zephyr split

dbingi-amd:
misc_props is SDT metadata from pcw.dtsi. Bare-metal domain generation must keep it so baremetal_xparameters_xlnx and bmcmake_metadata_xlnx can emit the corresponding macros and CMake variables. Linux and Zephyr have no binding for the node and should drop it.

When Zephyr handling moved out of gen_domain_dts, the original guard
    if (linux_dt or zephyr_dt) and node.name == "misc_props"
became an unconditional delete. That removed misc_props from the bare-metal prune as well, so the xparameters/CMake properties never appeared.

Restore the linux_dt-only delete in gen_domain_dts, and drop the node in zephyr_domain_dts after the shared domain prune.

### DIFF
--- a/lopper/assists/gen_domain_dts.py
+++ b/lopper/assists/gen_domain_dts.py
@@ -404,7 +404,7 @@ def xlnx_generate_domain_dts(tgt_node, sdt, options):
 
     mapped_children_nodes = []
     for node in root_sub_nodes:
-        if node.name == "misc_props":
+        if linux_dt and node.name == "misc_props":
             sdt.tree.delete(node)
             continue
         if linux_dt:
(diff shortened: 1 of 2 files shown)


dbingi-amd:
@sathishkumar-amd please review

sathishkumar-amd (approved):
The changes look good to me.

dbingi-amd:
@zeddii: Could you please review the changes

CC: @kedareswararao @sivadur

zeddii:
Confirmed the regression and the fix. `261b5563` dropped the `linux_dt` term along with the `zephyr_dt` one, and this puts it back.

Merged, since it's a live breakage in bare-metal generation and the fix is right. Please follow up with a test though. The failure mode here was the silent kind: the xparameters and CMake macros simply stopped appearing, with nothing to say why. `tests/test_xlnx_gen_domain.py` already builds trees this way, so a case asserting `misc_props` survives the bare-metal prune and is dropped for Zephyr is cheap, and would pin both halves of this change.

PR #836, opened 2026-09-18
lopper: assists: yaml_bindings: Keep xlnx,hdmi-mode for HDMI TX

gugulot:
## Summary
- Keep sdtgen `C_HDMI_MODE` (`xlnx,hdmi-mode`) in the Linux DT generated by `gen_domain_dts linux_dt`.
- That assist only retains properties listed in yaml_bindings `required:`.

## Test plan
- [x] Regenerated SDT from hw_hdmi21_tmdsonly_zcu102_0608.xsa
- [x] `gen_domain_dts psu_cortexa53_0 linux_dt` keeps `xlnx,hdmi-mode = <0x0>` on `v_hdmi_txss1`


zeddii:
Thanks for the patch — the change itself looks right, and adding the property to `required:` is the correct mechanism since that list is what the assist uses as its keep-list.

I've edited the description slightly to drop an internal reference. Worth keeping in mind for anything landing here: this repo is public, so ticket IDs and internal tooling or flow names shouldn't appear in PR descriptions, commit messages, or comments. Descriptions can be edited after the fact, but commit messages are permanent once merged, so those are the ones to watch.

Nothing to redo on your side — I'll pick this up from here.

zeddii:
Applied to master as `4d90960a`.

I amended the commit message slightly — one phrase reworded so it reads without needing internal context — but the change itself is untouched and your authorship and sign-off are preserved. Full suite passes.

Closing this since it's landed. Thanks for the patch.

PR #837, opened 2026-09-18
Add Versal SysMon support

dileepkumarnagavarapu:
This patch series adds support for AMD/Xilinx Versal SysMon in Lopper.

include the "xlnx,versal-sysmon" compatible entry, allowing proper
recognition of the SysMon node during device tree processing.

Versal SysMon nodes and generate appropriate domain-specific DTS
entries.

Tested on the vck190 platform with Lopper-generated dts

### DIFF
--- a/lopper/assists/zephyr_domain_dts.py
+++ b/lopper/assists/zephyr_domain_dts.py
@@ -750,6 +750,59 @@ def xlnx_remove_unsupported_nodes(tgt_node, sdt, machine, options=None):
                             node['power-delay-ms'] = 10
                         node.add(new_node)
                         node["compatible"] = "xlnx,versal-8.9a"
+                    # SYSMON
+                    if "xlnx,versal-sysmon" in node["compatible"].value:
+                        supply_count = sum(
+                            1 for child in node.child_nodes.values()
+                            if not child.props("xlnx,aie-temp")
+                        )
+                        try:
+                            node.delete("xlnx,numchannels")
+                        except KeyError:
+                            pass
+                        num_prop = LopperProp("xlnx,numchannels")
+                        num_prop.value = [supply_count]
+                        node + num_prop
+                        used_names = set()
+                        for child in list(node.child_nodes.values()):
+                            # Drop AIE temp pseudo-node (not part of SysMon supply child-binding)
+                            if child.props("xlnx,aie-temp"):
+                                try:
+                                    node.delete(child)
+                                except KeyError:
+                                    pass
+                                continue
+                            reg_val = child.propval("reg", list)
+                            if not reg_val or reg_val == ['']:
+                                try:
+                                    node.delete(child)
+                                except KeyError:
+                                    pass
+                                continue
+                            try:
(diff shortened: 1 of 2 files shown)


zeddii:
Applied to master as `8169b2e8`. The change itself is unmodified.

Two notes on what I adjusted and why.

I took only the third commit. The first two were already on master: they went in with #815, but that merge rebased them, so they landed as `b498f4fd` and `64420e00` rather than the SHAs on your branch. Git can't tell they're the same work, which is what made this show as conflicting. Worth knowing for next time: branching from current master rather than from the previous PR branch avoids it.

That also made the `Fixes:` tag point at `ba3fb18`, which doesn't exist upstream, so I repointed it at `64420e00` ... the same commit, post-rebase. It resolves now.

One observation for a follow-up rather than anything needing action here. `node.props("interrupts")` isn't an exact-name lookup: it tries the exact key first, then falls back to a regex over all property names, so it also matches `interrupts-extended`. A sysmon node carrying only `interrupts-extended` would therefore be treated as having interrupts and survive, which is the shape this commit exists to drop if Zephyr's binding wants `interrupts` specifically. Not reachable today since nothing in tree emits `interrupts-extended` on a sysmon node,  but `node.propval("interrupts") != ['']` would be the exact test if you want to close it off.

Closing this since it's landed. Thanks for the fix, and for the commit message. Naming the affected boards and why the driver can't cope made it easy to follow.

dileepkumarnagavarapu:
> Applied to master as `8169b2e8`. The change itself is unmodified.
> 
> Two notes on what I adjusted and why.
> 
> I took only the third commit. The first two were already on master: they went in with #815, but that merge rebased them, so they landed as `b498f4fd` and `64420e00` rather than the SHAs on your branch. Git can't tell they're the same work, which is what made this show as conflicting. Worth knowing for next time: branching from current master rather than from the previous PR branch avoids it.
> 
> That also made the `Fixes:` tag point at `ba3fb18`, which doesn't exist upstream, so I repointed it at `64420e00` ... the same commit, post-rebase. It resolves now.
> 
> One observation for a follow-up rather than anything needing action here. `node.props("interrupts")` isn't an exact-name lookup: it tries the exact key first, then falls back to a regex over all property names, so it also matches `interrupts-extended`. A sysmon node carrying only `interrupts-extended` would therefore be treated as having interrupts and survive, which is the shape this commit exists to drop if Zephyr's binding wants `interrupts` specifically. Not reachable today since nothing in tree emits `interrupts-extended` on a sysmon node, but `node.propval("interrupts") != ['']` would be the exact test if you want to close it off.
> 
> Closing this since it's landed. Thanks for the fix, and for the commit message. Naming the affected boards and why the driver can't cope made it easy to follow.

Thanks for the detailed explanation and for getting this merged.
 I'll make sure to base future work on current master to avoid that situation.
 Also, thanks for the note about node.props("interrupts") matching interrupts-extended. Currently, no SysMon nodes emit interrupts-extended, so this case isn't reachable today. I'll keep it in mind if support for such nodes is added in the future.

PR #838, opened 2026-09-18
Openamp xlnx update reg access

bentheredonethat:
use existing lopper utilities to access DT node 'reg' properties


zeddii (commented):
The cell-width work is right. `reg[1]`/`reg[3]` only ever worked for a two-address-cell, two-size-cell parent, and `_required_reg_region` puts that assumption in one place with a real error instead of an IndexError.

The `SHARED_MEM_SIZE` change deserves to be called out in the commit message as a bug fix rather than a cleanup. It was not just fragile, it was wrong. Line 693 on master reads `reg[1]` for the vdev0buffer size, which on a 2/2 layout is the low word of the address. Against the checked-in ZynqMP YAML, where `rpu0vdev0buffer` is `start: 0x9868000, size: 0x40000`, the generated header has been carrying:

```
#define SHARED_MEM_SIZE  0x9868000    <- the base address
```

The same function uses `reg[3]` for the vring sizes a few lines earlier, so the two are inconsistent with each other in the same block.

"Take SHARED_MEM_SIZE from the vdev0buffer region size" reads as a detail of the refactor. Anyone who has consumed a generated header has had a wrong value there, and that is the sort of thing people go looking for in the log afterwards. Same for `SHARED_BUF_OFFSET`: `vring_total_sz` and `shbuf_base - shm_base` agree only when the buffer starts immediately after the last vring, so where they diverged that was wrong output too.

The new `_error` on a vdev0buffer that precedes the vrings, and the one-buffer check, both look like the right checks for deriving an offset that way.

Two smaller things:

- `from lopper.assists.lopper_lib import ...` is an absolute package import in a file that otherwise reaches siblings by bare name (`from baremetalconfig_xlnx import get_cpu_node`). Both resolve; worth matching what is already there.
- `_required_reg_region` rejects `size <= 0`, which the elfload path did not check before. Stricter is better, but it is a behaviour change: a zero-size elfload carveout now fails where it previously returned a zero size and carried on.

The YAML mask migration in the second commit follows from the resolver work, and keeping the legacy 0x2 plus `cluster_cpu` form alive in the test rather than in the checked-in policies is the right way round.

bentheredonethat:
@zeddii  reworked

